PR #1, opened 2024-03-01
RGB Backlight: Dont update while screen is off and other improvements

HaxSam:
# What's new

- when the screen turns dark stop updating the rgb rainbow
- rgb rainbow mode has is own function and brightness change has is own function

-----
# For the reviewer


WillyJL:
can you test @Sil333033 ?

PR #2, opened 2024-03-01
Add is_connected api for expansion \w @Willy-JL

HaxSam:
# What's new

- Added new API to expansion to get the information if an expansion is initialized
- The expansion_worker communicates back to the control if there is an RPC connection
- new state for expansion for an established connection

@Willy-JL was heavily involved

-----
# For the reviewer


HaxSam:
Still need to add a better comment in `expension.h` but is nicer to view the changes here to see if I overlooked something at code review.

PR #3, opened 2024-03-01
Docs: Fix names in asset packs info

Mr-Proxy-source:
# What's new

- Changed link
- Changed name
-----
# For the reviewer


WillyJL:
good catch on the asset packs documentation! i was planning to redo it eventually so didnt pay too much attention to it, just added it because its needed.

for the mfc dict, they were contributed by third parties to xtreme at the time, would be wrong to pretend thats not the case imo.

Mr-Proxy-source:
Bet

PR #6, opened 2024-03-01
Expansion: Add is_connected API to check for VGM

HaxSam:
# What's new

- Added new API to expansion to get the information if an expansion is initialized
- The expansion_worker communicates back to the control if there is an RPC connection
- new state for expansion for an established connection

@Willy-JL was heavily involved

-----
# For the reviewer


WillyJL:
nice :D just made a few changes to naming and fixed some race conditions, thanks!

PR #7, opened 2024-03-03
Passport: Move text for better display

Kuronons:
Moving NAME, MOOD, LVL & XP/XP 1 pixel on the right for better display.


WillyJL:
LGTM, thx :D

PR #9, opened 2024-03-03
Little changes

grugnoymeme:
# What's new

- [Describe changes here]

-----
# For the reviewer


WillyJL:
The file names are mentioned to allow migration from xtreme firmware to momentum firmware. It makes a copy from the xtreme configs to the momentum configs, ensuring a smooth transition.

The link in the mf classic dict is to preserve origin. Those keys were submitted by users to the xtreme project, would be wrong to erase that fact.

Thanks but this should not be merged

PR #16, opened 2024-03-07
JS : Add GoodUSB script for Windows 

xAstroBoy:
# What's new

Add a small GoodUSB script for windows Builds

-----
# For the reviewer


WillyJL:
i dont think this belongs in the base firmware. i see youve already shared in our discord server, maybe contribute it to another file dump like uberguidoz's or similar...

thank you regardless!

PR #17, opened 2024-03-07
JS: Add `parse_int()`, `to_upper_case()`, `to_lower_case()`

Spooks4576:
# What's new

Added a Global StringUtils Library that gets handled globally also added a ParseInt global function similar to what regular js has

-----
# For the reviewer


WillyJL:
as per [mjs docs](https://github.com/cesanta/mjs):
- `substring("abc", 1, 3)` -> `"abc".slice(1, 3)`
- `slice("abc", 1, 3)` -> `"abc".slice(1, 3)`
- `indexOf("abc", "bc")` -> `"abc".indexOf("bc")`
- `GetLength("abc")` -> `"abc".length`

as for `parse_int("123")`, mostly ok, just some tiny changes to error handling, nothing major, well done

`toUpperCase("abc")` and `toLowerCase("ABC")` i'd like to have on the string objects themselves. i will see if this can be done easily, if not can keep global (though i will move into `js_thread.c`, having a separate header which actually contains code, and is labelled as a module when its global, is not ideal.

also, please stick to a naming convention. `parseint`, `parse_int`, `indexOf` and `GetLength` can't possibly all belong to the same API... considering the existing examples from OFW js runner, i'd say the safe bet is `snake_case` globals, and `camelCase` module functions.

WillyJL:
adding to string object as `"abc".toUpperCase()` would go into `getprop_builtin_string()` of `lib/mjs/mjs_exec.c`, which means it would end up in firmware flash. i'd rather have it in js_app, so its in fap instead. ill make it respect the global function naming scheme as `to_upper_case("abc")`

WillyJL (approved):

PR #29, opened 2024-03-13
Adding Yappy profile picts for MNTM asset pack

Kuronons:
Happy / okay / bad
.png


WillyJL:
I find it funny that the happy one almost looks more indifferent than the okay one, but that's normal for capybaras xD

Thank you so much! <3 they look great and adorable

PR #38, opened 2024-03-16
Docs: Added JS API Documentation

freehuntx:
# What's new
I created an documentation for the JavaScript api. Its not finished yet and still WIP but i think any documentation is better than none. This makes sure we already have a place for further documentation and we can give people the link to the doc instead of telling them to look in the code.

- Created file: `documentation/JavaScript.md`

-----
# For the reviewer
Its still WIP but all methods are already listed. So its still useful.

### DIFF
--- a/documentation/JavaScript.md
+++ b/documentation/JavaScript.md
@@ -1,0 +1,148 @@
+
+# JavaScript scripting API (WIP)
+## Note: This documentation is still work in progress! 
+Todo:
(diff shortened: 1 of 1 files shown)


WillyJL (approved):

WillyJL:
thanks! will keep adding in the future, also possibly specifying what is in OFW and what isn't, but thisis a great start!

PR #47, opened 2024-03-19
SubGHz: Add back region lock and manual bypass option

Sil333033:
# What's new

- Functional region hal back to the Flipper. Now we can use the regions to determine if you are allowed to transmit on the chosen frequency. 
- Brought back an option to bypass it inside the Momentum app.
- Reworked 'furi_hal_subghz_is_tx_allowed()' to actually use the region check.
- Flipper gets forced rebooted after changing the SubGHz region bypass or frequency extenstion inside the Momentum app, because those settings are loaded on startup. We could change it with an extra 'furi_hal_subghz_set_extended_frequency(bool state)' function, but these settings aren't changed often.

-----
# For the reviewer

### DIFF
--- a/applications/main/momentum_app/momentum_app.c
+++ b/applications/main/momentum_app/momentum_app.c
@@ -292,14 +292,14 @@ MomentumApp* momentum_app_alloc() {
 
         if(!flipper_format_rewind(file)) break;
         while(flipper_format_read_uint32(file, "Frequency", &temp, 1)) {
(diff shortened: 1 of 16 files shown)


WillyJL (approved):

WillyJL:
> We could change it with an extra 'furi_hal_subghz_set_extended_frequency(bool state)' function, but these settings aren't changed often.

I wanted to do this, but actually it would require exposing this functionality to API symbols... not worth it :D

PR #53, opened 2024-03-24
VGM/RPC: Rework custom colors, smoother config, better rainbow support

HaxSam:
# What's new

- would make a smother update of rgb color to the vgm

-----
# For the reviewer


zacharyweiss:
A suggestion, if acceptable computational- / memory-intensity-wise (might be worthy of a separate PR, so apologies if this comment is misplaced). Interpolate over a perceptually uniform colorspace a la [OKLAB](https://bottosson.github.io/posts/oklab/) instead of over the RGB/HSV values, as the latter (the current approach) leads to perceptual peaks/valleys in lightness. On-the-fly conversion requires some floating-pt math (might be intensive for what's intended as a background process?); alternatively, lookup tables could be precomputed for speed, if there's sufficient memory.

Code for conversion to/from linear sRGB listed in the above-linked post; would need minor modification / composition with sRGB<>RGB565 for conversion to/from RGB565:

```
struct Lab {float L; float a; float b;};
struct RGB {float r; float g; float b;};

Lab linear_srgb_to_oklab(RGB c) {
    float l = 0.4122214708f * c.r + 0.5363325363f * c.g + 0.0514459929f * c.b;
    float m = 0.2119034982f * c.r + 0.6806995451f * c.g + 0.1073969566f * c.b;
    float s = 0.0883024619f * c.r + 0.2817188376f * c.g + 0.6299787005f * c.b;

    float l_ = cbrtf(l);
    float m_ = cbrtf(m);
    float s_ = cbrtf(s);

    return {
        0.2104542553f*l_ + 0.7936177850f*m_ - 0.0040720468f*s_,
        1.9779984951f*l_ - 2.4285922050f*m_ + 0.4505937099f*s_,
        0.0259040371f*l_ + 0.7827717662f*m_ - 0.8086757660f*s_,
    };
}

RGB oklab_to_linear_srgb(Lab c) {
    float l_ = c.L + 0.3963377774f * c.a + 0.2158037573f * c.b;
    float m_ = c.L - 0.1055613458f * c.a - 0.0638541728f * c.b;
    float s_ = c.L - 0.0894841775f * c.a - 1.2914855480f * c.b;

    float l = l_*l_*l_;
    float m = m_*m_*m_;
    float s = s_*s_*s_;

    return {
        +4.0767416621f * l - 3.3077115913f * m + 0.2309699292f * s,
        -1.2684380046f * l + 2.6097574011f * m - 0.3413193965f * s,
        -0.0041960863f * l - 0.7034186147f * m + 1.7076147010f * s,
    };
}
```

lib/momentum/rpc_rgb.c
```diff
@@ -0,0 +1 @@
+#include "rpc_rgb.h"
```
WillyJL: are these 2 needed?

assets/protobuf
WillyJL, line 1: if this is now a fork, might need to update .gitmodules

applications/main/momentum_app/scenes/momentum_app_scene_misc_vgm.c
```diff
@@ -31,58 +38,73 @@ static void momentum_app_scene_misc_vgm_colors_changed(VariableItem* item) {
...
-    {"Blue", {0x001F}},
-    {"Navy", {0x0010}},
-    {"Purple", {0x8010}},
-    {"Fuchsia", {0xF81F}},
-    {"Pink", {0xA8F5}},
-    {"Brown", {0xA145}},
-    {"White", {0xFFFF}},
+    {"Off", {{0, 0, 0}}},
+    {"Rainbow", {{0, 0, 0}}},
+    {"RgbMod", {{0, 0, 0}}},
```
WillyJL, line 20: is the intent to set rainbow and rgbmod modes with the foreground and background toggles? if so, we can get rid of the other entry in the menu for "VGM Colors"

lib/momentum/momentum.h
```diff
@@ -88,9 +96,9 @@ typedef struct {
     FuriHalSerialId uart_nmea_channel;
     bool file_naming_prefix_after;
     VgmColorMode vgm_color_mode;
-    Rgb565Color vgm_color_fg;
-    Rgb565Color vgm_color_bg;
     FuriHalVersionColor spoof_color;
+    RgbColorTransmit vgm_color_fg;
+    RgbColorTransmit vgm_color_bg;
```
WillyJL: changing data type here, might be good to change name too so it is not imported incorrectly. then again, also need to update the code part in settings.c, so maybe can change it to save as hex and check data length instead of saving as uint32

WillyJL (changes requested):

lib/momentum/rpc_rgb.c
HaxSam: Not anymore I could delete those files

applications/main/momentum_app/scenes/momentum_app_scene_misc_vgm.c
HaxSam: Yes because you can now set the background or the foreground with rainbow and not one of them.
Wanted to ask if we should restructure or keep it like how I did it.

WillyJL (approved):
LGTM :D

PR #55, opened 2024-03-27
JS: Add __filepath and __dirpath globals.

jamisonderek:
# What's new

- This adds two new global variables to JavaScript files loaded by the js_app.  `__filepath` is the fully qualified path to the .js file (like "/ext/apps/Scripts/path.js").  `__dirpath` is the folder containing the .js file (like "/ext/apps/Scripts").

-----
# For the reviewer


HaxSam:
doing the review

HaxSam:
Looks good

WillyJL:
Thank you :D

PR #56, opened 2024-03-27
JS: Add widget module

jamisonderek:
# What's new

- Add widget module.  The widget module has a bunch of addXXX methods for adding frame, rframe, box, rbox, circle, disc, dot, glyph, icon, line, text and xbm images.  Every time you call add, it returns an id that can be used by remove if you want to remove the item later. show() is non-blocking UI, so your JS can do other processing while the UI is displayed. The user can click the back button or you can call close() to dismiss the view.

-----
# For the reviewer


applications/system/js_app/modules/js_widget.c
```diff
@@ -0,0 +1,1019 @@
...
+    const char* icon_name = mjs_get_string(mjs, &icon_arg, &icon_len);
+    if(!icon_name) {
+        ret_bad_args(mjs, "Icon name must be a string");
+        return;
+    }
+
+    const Icon* icon = NULL;
+    for(size_t i = 0; i < ICON_PATHS_COUNT; i++) {
+        if(ICON_PATHS[i].path != NULL && strnlen(ICON_PATHS[i].path, icon_len + 2) == icon_len &&
+           strncmp(icon_name, ICON_PATHS[i].path, icon_len) == 0) {
```
WillyJL: Ideally, it would be best to use the same names as apps do in the firmware, which would be without the path before, the `ICON_PATHS` array includes the path due to finding the icon file on sd card for asset packs, here we could use `strrchr()` to find the last `/` and compare after that (the icon paths are guaranteed to have atleast one folder, so `strrchr()` will always return a result)

WillyJL:
Aside from that single nitpick, this is looking great, thank you so much! This will prove very useful for many scripts.

I can get that small change done tomorrow unless you do it before me 😄

applications/system/js_app/modules/js_widget.c
WillyJL: I'm working on this right now

WillyJL: done

WillyJL (approved):

PR #58, opened 2024-03-28
FindMy Add Ability to Show MAC

MatthewKuKanich:
# What's new

- Added show_mac toggle to save state.
- Reworked GUI logic and formatting
- Added not just a Yes option but also a No option ;)

-----
# For the reviewer

- [X] I've uploaded the firmware with this patch to a device and verified its functionality
- [X] I've confirmed the bug to be fixed / feature to be stable


WillyJL (approved):
LGTM :D

PR #62, opened 2024-03-31
NFC MBTA CharlieCard parsing plugin

zacharyweiss:
# What's new

Built NFC plugin for MBTA CharlieCard (Boston, MA, USA)
- Implements `read`, `verify`, and `parse`
- Lookups for all card types, and Blue & Orange Line fare gate IDs
- Passes, some trip flags, and misc other fields yet to be reverse engineered

I'm not particularly adept with C, nor experienced in contributing to open-source projects, so feedback / review of the following sought, in decreasing order of importance:
- Let me know if this is better suited as a PR to OFW, and allowed to percolate up to the CFWs. Initial PR made here simply as this is the FW I'm currently running (this plugin does not rely on any FW-specific features, to my knowledge); happy to see it included in any FWs, so long as credit & license is kept.
- Check for memory leaks — I likely am not following best practices with memory management & cleanup, due to my ignorance of C.
- Check that I am not being overly greedy with card validation & producing false-positives parses; the `read` and `verify` functions are largely taken as boilerplate from analogous plugins, as are the initial checks at the start of the `parse` function.
- For those in Boston: help build out the mapping of fare gate IDs to stations! Take note of the IDs printed in the transaction log, and associate them to a station in the lengthy fare gate struct.

Credit & citations listed in header comment.

-----
# For the reviewer

### DIFF
--- a/applications/main/nfc/plugins/supported_cards/charliecard.c
+++ b/applications/main/nfc/plugins/supported_cards/charliecard.c
@@ -1,0 +1,1032 @@
+/*
+ * Parser for MBTA CharlieCard (Boston, MA, USA).
+ *
+ * Copyright 2024 Zachary Weiss <me@zachary.ws>
+ * 
+ * Public security research on the MBTA's fare system stretches back to 2008,
+ * starting with Russel Ryan, Zack Anderson, and Alessandro Chiesa's 
+ * "Anatomy of a Subway Hack", for which they were famously issued a gag order. 
+ * A thorough history of research & researchers deserving of credit is 
+ * detailed by @bobbyrsec in his 2022 blog post (& presentation):
+ * "Operation Charlie: Hacking the MBTA CharlieCard from 2008 to Present" 
+ * https://medium.com/@bobbyrsec/operation-charlie-hacking-the-mbta-charliecard-from-2008-to-present-24ea9f0aaa38
+ * 
+ * Fare gate IDs, card types, and general assistance courtesy of the
+ * minds behind DEFCON 31's "Boston Infinite Money Glitch" presentation:
+ * — Matthew Harris; mattyharris.net <matty@mattyharris.net>
+ * — Zachary Bertocchi; zackbertocchi.com <zach@zachbertocci.com>
+ * — Scott Campbell; josephscottcampbell.com <scott@josephscottcampbell.com>
+ * — Noah Gibson; <noahgibson06@proton.me>
+ * Talk available at: https://www.youtube.com/watch?v=1JT_lTfK69Q
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "nfc_supported_card_plugin.h"
+#include <flipper_application.h>
+
+#include <nfc/protocols/mf_classic/mf_classic_poller_sync.h>
+
+#include <bit_lib.h>
+#include <datetime.h>
+#include <furi_hal.h>
+#include <locale/locale.h>
+
+#define TAG "CharlieCard"
+
+// starts Wednesday 2003/1/1 @ midnight
+#define CHARLIE_EPOCH          \
+    (DateTime) {               \
+        0, 0, 0, 1, 1, 2003, 4 \
+    }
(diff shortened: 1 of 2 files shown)


applications/main/nfc/plugins/supported_cards/charliecard.c
```diff
@@ -0,0 +1,1032 @@
...
+        furi_string_cat_printf(parsed_data, "\nExpired: %s", expired(e_v, last) ? "Yes" : "No");
+
+        const Trip* trips = trips_parse(data);
+        furi_string_cat_printf(parsed_data, "\nTransactions:");
+        for(size_t i = 0; i < CHARLIE_N_TRIP_HISTORY; i++) {
+            furi_string_cat_printf(parsed_data, "\n");
+            trip_format_cat(parsed_data, trips[i]);
+            furi_string_cat_printf(parsed_data, "\n");
+        }
+        // free(trips);
```
WillyJL: this is the only issue i can see, memory-management-wise. `trips_parse()` allocates the buffer, it is used here, and then it's not free'd. simply uncommenting that line should work, is there a particular reason why it was commented?

WillyJL (changes requested):

WillyJL:
to respond in order:
- absolutely, i suggest proposing a PR to OFW aswell! we welcome the changes here and usually they will get accepted faster into CFW's, but it would be great if it reached more users. can't speak as to whether they would accept it in OFW, but i'd say it's definitely worth a shot. also, it's an honor to hear you're using Momentum :D
- from what i can see the code looks quite solid and clean, well done! only the one missing `free()` as i mentioned in the review, but other than that this is great as far as i'm concerned
- i'm not that experienced with NFC plugins in that sense, i will get some smarter friends than me to have a look ;) but atleast, the read function looks solid, checking for error none rather than assuming a full read, handles partial reads correctly (this was a wide-spread issue a little while back)

thank you!

applications/main/nfc/plugins/supported_cards/charliecard.c
zacharyweiss: Ope, good catch! I had commented that out as it wasn't compiling with it uncommented, and simply forgot to go back and check why I was having compile-time issues to begin with. Turns out I simply had an unnecessary `const` on line 999: `const Trip* trips = trips_parse(data);` [sic]. Removing the `const` and uncommenting the `free()` makes the compile-time checks happy, and should fix the above.

applications/main/nfc/plugins/supported_cards/charliecard.c
```diff
@@ -833,6 +863,9 @@ static uint16_t type_parse(const MfClassicData* data) {
     // bitshift (2bytes = 16 bits) by 6bits for just first 10bits
     const uint16_t type1 = pos_to_num(data, 2, 1, 0, 2) >> 6;
     const uint16_t type2 = pos_to_num(data, 3, 1, 0, 2) >> 6;
+    // might be wise to remove the assertion; then again, it's an effective way
+    // to crowdsource research, as hopefully if this isn't universally true,
+    // someone will come running when their app crashes — probably not a best practice though haha.
     furi_assert(type1 == type2);
```
WillyJL, line 865: furi_assert() is compiled only when DEBUG=1 is enabled, which it is not by default, nor for production builds in Momentum, you'd need to compile with `./fbt <command> DEBUG=1`, so assert should be fine here! if you want to enforce it in all builds, regardless of DEBUG flag, you can use `furi_check()`

WillyJL:
I asked @Leptopt1los from unleashed team about card verification, he said it looks great!

The only thing he advised is maybe not only checking key A but also the paired key B in parse(), just to be safe.

If you want to ge ahead with that, by all means do, otherwise we can merge as is :D

applications/main/nfc/plugins/supported_cards/charliecard.c
zacharyweiss: Good to know! For my own education: does that mean compiling without DEBUG strips that, and is independent of the debug setting on the flipper? Or is the check kept in the compiled fap regardless, and the DEBUG compile arg simply sets the debug mode flag (locally or globally) on the flipper after flashing/launching?

zacharyweiss:
Key B check added! Should be all set now. Time permitting, I'll also likely be slowly digging into the few remaining unknowns on the card over the coming weeks; is there a preferred PR / contributing etiquette here? (ie, only PR at "project end" to minimize the number of PRs to review? PR with each individual discovery / improvement to keep things bleeding-edge? other?)

Thanks again for y'all's input & help!

applications/main/nfc/plugins/supported_cards/charliecard.c
WillyJL: `DEBUG=1` refers to the compiler options, with `DEBUG=0` (default) `furi_assert()` will be stripped completely. the `Debug` option in flipper settings allows to attacha live debugger, and to access some extra hidden functionality (raw reads, debug info in some parsers...), but will not bring back the stripped `furi_assert()` calls

WillyJL:
perfect, thank you! yes, feel free to make new PRs as you progress with the parser. as for timing, whatever works best for you, as long as its not like once per day xD

WillyJL (approved):

PR #64, opened 2024-04-01
MNTM Pack - Adding U2F + Settings Yappy picts set

Kuronons:
O_oV

B&W pngs. No idea why originals were black/orange...


WillyJL (approved):
Awwwww

PR #65, opened 2024-04-01
MNTM Pack - Adding NFC emulation & IR remote recording graphic assets

Kuronons:
Yap²

In addtition to yappy insert, edited the IR remote's Flip so it fully displays on screen. (was originally cut on the right side)

Note that NFC pict is 64px high... and is not set properly on y=0. 
Managed to fit the drawn part on Flip screen but this could (should?) be code-edited imo... no reason a pict is set to partially display out of the screen.


WillyJL:
> Note that NFC pict is 64px high... and is not set properly on y=0.
Managed to fit the drawn part on Flip screen but this could (should?) be code-edited imo... no reason a pict is set to partially display out of the screen

ironically, some of the uses of `NFC_dolphin_emulation_51x64` are inside the lfrfid app and are correctly at y=0, theres only on reference in the nfc app, and thats at y=3. i think it should be fine, i get what you mean, but also the default icon touches the edge of the screen at the top, so i guess it was on purpose to have it shifted a bit

thank you! :D

WillyJL (approved):

PR #72, opened 2024-04-04
MNTM Pack - Erase Pin yappy asset

Kuronons:
Is that asset really used ? Couldn't find one menu where it showed up. (Haven t looked in the code tho...) 
Anyway, was fun to make. o_O;


WillyJL (changes requested):
needs to be in assets/icons, asset packs aren't loaded in recovery mode

WillyJL:
> Is that asset really used ? Couldn't find one menu where it showed up.

yep, try rebooting with back+left, then before it starts booting hold up (basically, if youre holding up while booting, it triggers). the slower but easier way is holding back+up for 30 seconds

WillyJL (approved):
:heart_eyes:

PR #73, opened 2024-04-04
Replacing stock DFU graphic asset by yappy one

Kuronons:
O_oV


WillyJL (approved):
Poor lil fella, so scared xD

PR #80, opened 2024-04-06
Storage: USB Mass Storage enhancement for setting volume label

jetblk:
# What's new

- Previously the default USB Mass Storage label was unset ("NO NAME"). Now, the label is set to the Image Name the user has input when the storage is created.

-----
# For the reviewer

- [x] I've uploaded the firmware with this patch to a device and verified its functionality. Verified with all the latest dev changes as of April 6, 2024.
- [x] I've confirmed the bug to be fixed / feature to be stable. Created multiple volumes with a name and without (size gets set as name, thus the volume name as well).

### DIFF
--- a/applications/services/storage/storages/storage_ext.c
+++ b/applications/services/storage/storages/storage_ext.c
@@ -5,6 +5,7 @@
 #include <furi_hal.h>
 #include "sd_notify.h"
 #include <furi_hal_sd.h>
+#include <toolbox/path.h>
 
 typedef FIL SDFile;
 typedef DIR SDDir;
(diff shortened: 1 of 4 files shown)


applications/services/storage/storages/storage_ext.c
```diff
@@ -739,6 +740,13 @@ FS_Error storage_process_virtual_format(StorageData* storage) {
     SDData* sd_data = storage->data;
     uint8_t* work = malloc(_MAX_SS);
     SDError error = f_mkfs(sd_data->path, FM_ANY, 0, work, _MAX_SS);
+    storage_process_virtual_mount(storage);
+    const char* path = storage_file_get_path(mnt_image, mnt_image_storage);
+    char* label = basename(path);
+    int len = strlen(label);
+    label[len - 4] = '\0'; // truncate the .img extension
```
WillyJL: This is effectively editing the memory of the furi string directly, which should never be done. You can malloc a smaller buffer of the appropriate size to get just filename without extension, and memcpy into it, then set the null terminator on that copied buffer

WillyJL (changes requested):

applications/services/storage/storages/storage_ext.c
jetblk: Thanks for the feedback, will modify!

WillyJL: Or use strlcpy with the actual buffer size (including the null terminator), which will figure it out appropriately

jetblk: Changes submitted. Thanks for the guidance, appreciate the feedback!

WillyJL (approved):

PR #98, opened 2024-04-17
MNTM FW - Replacing NO SD internal anim by custom1-frame one

Kuronons:
Saving octets by getting rid of 5 frames out of 6.
-> Static "anim". Yappy themed.
Got rid of the "Need an SD card" bubble and replaced it by drawn "Insert SD card" text on png.


WillyJL (approved):
thanks! will check exactly how much flash it saves

PR #100, opened 2024-04-19
FindMy Flipper Battery Status Refactor, Unify State Sync

MatthewKuKanich:
# What's new

- [FindMy Flipper Battery Status Refactor, Unify State Sync]

-----
# For the reviewer


applications/system/findmy/findmy_state.c
```diff
@@ -42,6 +42,13 @@ bool findmy_state_load(FindMyState* out_state) {
             }
             state.tag_type = tmp;
 
+            if(!flipper_format_read_uint32(file, "battery_level", &tmp, 1)) {
+                tmp = 0x00; // Default battery level set to Full
+                flipper_format_rewind(file);
+            }
+            FURI_LOG_I("findmy_load", "Saved Battery: %ld", tmp);
+            state.battery_level = tmp;
```
WillyJL: is it really necessary to save the battery level in the config? if youre gonna update the data with the current battery level at runtime, and its stored in the data anyway, this seems redundant

applications/system/findmy/findmy_i.h
```diff
@@ -38,6 +38,7 @@ struct FindMy {
 
     uint8_t mac_buf[EXTRA_BEACON_MAC_ADDR_SIZE];
     uint8_t packet_buf[EXTRA_BEACON_MAX_DATA_SIZE];
+    uint8_t battery_level;
```
WillyJL: i believe this is leftover, youre keeping batter_level in the state struct no?

WillyJL (changes requested):
minor nitpicks, you know me )

applications/system/findmy/findmy_state.c
MatthewKuKanich: Not necessary, and I removed it for a little while. I added it back because I liked having it saved for debugging reasons but can be removed.

WillyJL (approved):
LGTM, thanks for bearing with me )

PR #103, opened 2024-04-19
JS: Fix widget sometimes MemCrash on exit.

jamisonderek:
# What's new

- Bug fix where sometimes JavaScript scripts were crashing on exit if they used the JS widget feature.

-----
# For the reviewer


WillyJL (approved):
LGTM, thanks!

PR #104, opened 2024-04-21
Fix MNTM-DEV git hash formatting

zacharyweiss:
# What's new

For a (currently unknown) reason, the built `version.inc.h` `#define GIT_COMMIT` on my [dev branch commit cacdc68](https://github.com/zacharyweiss/Momentum-Firmware/tree/cacdc6841afa1fc3f21767b42f75bbb22606ddb2) was populated with the first 8 characters of the commit hash when built, rather than the standard first 7 characters. As a result, when displayed in `momentum_app_scene_start` the hash header spills over the edge of the screen. 

While I don't diagnose what causes the oddity in # chars of the built `GIT_COMMIT`, this PR adds a simple max `strlen` of 7 for the displayed hash. `strlen` still checked, in the event there are fewer than 7 chars.

-----
# For the reviewer

### DIFF
--- a/applications/main/momentum_app/momentum_app.c
+++ b/applications/main/momentum_app/momentum_app.c
@@ -39,14 +39,12 @@ bool momentum_app_apply(MomentumApp* app) {
                    file, SUBGHZ_SETTING_FILE_TYPE, SUBGHZ_SETTING_FILE_VERSION))
                 break;
 
-            while(flipper_format_delete_key(file, "Add_standard_frequencies"))
-                ;
+            while(flipper_format_delete_key(file, "Add_standard_frequencies"));
             flipper_format_write_bool(
                 file, "Add_standard_frequencies", &app->subghz_use_defaults, 1);
 
             if(!flipper_format_rewind(file)) break;
-            while(flipper_format_delete_key(file, "Frequency"))
-                ;
+            while(flipper_format_delete_key(file, "Frequency"));
             FrequencyList_it(it, app->subghz_static_freqs);
             for(size_t i = 0; i < FrequencyList_size(app->subghz_static_freqs); i++) {
                 flipper_format_write_uint32(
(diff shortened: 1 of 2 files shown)


WillyJL:
Good catch. Still I'll try to find why it's 8 instead of 7. Honestly it might just be that we are starting to get ambiguous commit hashes? Sounds far fetched but maybe

zacharyweiss:
Definitely an odd bug; just popped this superficial fix on as it was bugging me this morning. Do you know where in the build tools / code those version defines get made? Haven't spent any time digging through `fbt` / the toolchain yet, so wasn't sure where to start looking for the root cause.

zacharyweiss:
> Honestly it might just be that we are starting to get ambiguous commit hashes?

Gotta start running [chosen prefix](https://eprint.iacr.org/2020/014.pdf) attacks to test how the system handles ambiguous truncated hashes lmao

zacharyweiss:
Found it. The script had 8 hardcoded as the length; must've just gotten lucky with the character widths up till now for the 8th character to be fully hidden offscreen (is the font here not fixed-width?).

WillyJL:
the font is not fixed width unfortunately, its the primary font used for titles and such.

considering length of 8 hardcoded in fbt, there could be an expectation that this be respected across different firmwares, so im not sure if changing it here without discussing with OFW would be a good idea, especially considering flipperdevices/flipperzero-firmware#2497, ill ask aku or zlo if they know anything

zacharyweiss:
Good thinking. At worst can leave as just the fix at the display step

WillyJL (approved):
LGTM! thanks

WillyJL:
for some context, heard back from aku that it always was 8, and at some point it became 7 on some systems so they fixed inconsistencies like that. ill double check the devbuild system and artifacts and decide what to do for those, but for the actual githash in firmware, that should stay 8 charatcers

WillyJL:
indeed, the CI build used 8 characters like that, while the local build used 7 since i made that and never noticed. atleast the devbuilds part doesnt need changes

PR #105, opened 2024-04-21
Rework MNTM menu, fix Long Dolphin Name, add Battery Charge %, add Ex…

MatthewKuKanich:
…t 5v Indicator

# What's new

- Changed position of dolphin name to title spot
- Added battery percentage info and a charging icon (When charging)
- Added an indicator for if external 5v is enabled (OTG)

-----
# For the reviewer


WillyJL (approved):
Looks nice )
thanks!

PR #110, opened 2024-04-26
Archive: Favorite .mag files

zacharyweiss:
# What's new

- Impl favoriting `.mag` files (for use with [magspoof commit >= 3ad981c](https://github.com/zacharyweiss/magspoof_flipper/commit/3ad981c8131acc117a39395f4dbbdb71f48d0e95))
- Adds `mag_card_10px.png` to `archive`;  `I_mag_card_10px` to symbols

-----
# For the reviewer


zacharyweiss:
Nix this if it seems out-of-scope, I can imagine that adding each EXT fap to have favoritable files could become unwieldy to maintain; don't want to set the wrong precedent by accident.

zacharyweiss:
Do I need to manually run the build with f18 targets to configure those symbols properly? Haven't made FW PRs that cause a change to the API ver before

targets/f7/api_symbols.csv
```diff
@@ -1,5 +1,5 @@
 entry,status,name,type,params
-Version,+,61.1,,
+Version,+,61.2,,
```
WillyJL: We need to keep our API version in sync with OFW. When we add or change something, API version should stay the same. Side note, icons being in API is exclusive to momentum and xfw due to how asset packs work

applications/main/archive/helpers/archive_browser.h
```diff
@@ -24,25 +24,16 @@ static const char* tab_default_paths[] = {
...
+    [ArchiveFileTypeIButton] = ".ibtn",     [ArchiveFileTypeNFC] = ".nfc",
+    [ArchiveFileTypeSubGhz] = ".sub",       [ArchiveFileTypeLFRFID] = ".rfid",
+    [ArchiveFileTypeInfrared] = ".ir",      [ArchiveFileTypeSubghzPlaylist] = ".txt",
+    [ArchiveFileTypeSubghzRemote] = ".txt", [ArchiveFileTypeInfraredRemote] = ".txt",
+    [ArchiveFileTypeBadKb] = ".txt",        [ArchiveFileTypeWAV] = ".wav",
+    [ArchiveFileTypeMag] = ".mag",          [ArchiveFileTypeU2f] = "?",
+    [ArchiveFileTypeApplication] = ".fap",  [ArchiveFileTypeJS] = ".js",
+    [ArchiveFileTypeSearch] = "*",          [ArchiveFileTypeUpdateManifest] = ".fuf",
+    [ArchiveFileTypeDiskImage] = ".img",    [ArchiveFileTypeFolder] = "?",
+    [ArchiveFileTypeUnknown] = "*",         [ArchiveFileTypeAppOrJs] = ".fap|.js",
```
WillyJL: I hate when it formats the arrays like this, will make it so annoying to merge future changes too. I don't think there's a consistent way to make it not do that, I'll try to fiddle with it a bit

applications/main/archive/scenes/archive_scene_browser.c
```diff
@@ -34,6 +34,8 @@ const char* archive_get_flipper_app_name(ArchiveFileTypeEnum file_type) {
         return "Bad KB";
     case ArchiveFileTypeWAV:
         return EXT_PATH("apps/Media/wav_player.fap");
+    case ArchiveFileTypeMag:
+        return EXT_PATH("apps/GPIO/magspoof.fap");
```
WillyJL, line 38: Also reminds me about the wav player, for the previous files we added support for, we moved the applications to the firmware repository. The idea was that if one day we were to make a build without apps included, these that have file associations would be included anyway. But now I'm not so sure, on both of these things. I'm not sure if we'd actually ever have a build without apps, sounds like unnecessary complexity and confusion, and would need some rework of the webupdater. And even if we do decide to do that, I'm not sure that including the apps regardless just because they have file associations, would make sense. Could just be that if you did install one without apps, then it would give an error for the app missing...

Nothing to change here, just a discussion we'll have to go over with the team

WillyJL (changes requested):

WillyJL:
I'll get the app updates merged in and update the submodule in this Pr

Also no, we don't deal with f18 (yet) so we leave it fully untouched to avoid unnecessary merge conflicts

targets/f7/api_symbols.csv
zacharyweiss: Gotcha gotcha, makes sense. Fine to manually decrement?

WillyJL: Yep!

applications/main/archive/helpers/archive_browser.h
zacharyweiss: Drives me up a wall too; it made one of the lists in my transit card parser practically unreadable. If there's an easy way to disable (and that won't cause issues a la failing lint on other FWs) I'd love to.

applications/main/archive/scenes/archive_scene_browser.c
zacharyweiss: Hm, I wonder if we could do something clever like implement an optional `ArchiveSupportedFaps` interface, like NFC does with `NfcSupportedCardsPlugin`, where one can populate the relevant stuct fields (just file icon and fap path?) and then the list gets auto-built onload of the archive. Probably out of scope, but would eliminate maintenance edits like this, and handle the above edge case you're mentioning nicely. Would leave it up to FAP authors to implement or not, and I don't think it'd cause conflicts across FWs (or if so, easy enough to just wrap in an `ifdef` like the other FW-specific features)

applications/main/archive/helpers/archive_browser.h
WillyJL: found it )

WillyJL (approved):
LGTM, works nicely. thanks!

WillyJL:
not sure why the build is failing tho

WillyJL:
ah its because of my changes to fbt_options.py, doesnt seem to play well here for some reason, probably due to how the repo is cloned on PRs

PR #113, opened 2024-05-03
MNTM Pack - Adding Yappy LEVEL UP "anim"

Kuronons:
And that is the final asset for this pack ! O_oV


WillyJL (approved):
coolest yap in the west :P

thanks!

PR #114, opened 2024-05-03
Updater: Gzipped resources (220% faster FW upload) + Updater improvements

WillyJL:
# What's new

- Include uzlib into firmware libraries (based on OFW PR 3158), not exposed to API (yet)
- Tar archive supports `.tar.gz` on-the-fly uncompression (based on OFW PR 3158 code for gzip CLI)
- Tar archive supports read progress callback (we don't know how many files are in tar due to being compressed, takes too long to count)
- Rebalance updater stages weights, progressbar is more consistent now
- Show stage progress percent in updater
- Gzip resources.tar (~12MB -> ~5MB), 220% faster upload and 14% slower unpack, at current time ~30s saved overall in FW install on fast USB, slower USB might save more

-----
# For the reviewer

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -38,3 +38,6 @@
 [submodule "lib/stm32wb_copro"]
 	path = lib/stm32wb_copro
 	url = https://github.com/flipperdevices/stm32wb_copro.git
+[submodule "lib/uzlib"]
+	path = lib/uzlib
+	url = https://github.com/pfalcon/uzlib.git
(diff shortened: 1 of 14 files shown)


WillyJL:
I heard from aku that zlo had some issues with uzlib, I'm yet to heard back from him. Seems to be working fine for me for now tho

WillyJL:
merging into dev so we can get more testing, should be mostly fine however

PR #117, opened 2024-05-03
Infrared: Add AC model Pioneer Inverter

djcas9:
# What's new

- Add AC model: Pioneer Inverter

-----
# For the reviewer


WillyJL (approved):
LGTM, thanks!

PR #119, opened 2024-05-06
Improve Hormann BiSecur readability in SubGHz history

user890104:
# What's new

- SubGHz decoders can provide a one-line string to be used in the history list.
- Improve Hormann BiSecur readability in SubGHz history, so that the serial number and a hash of the encrypted bytes can be seen without exiting the list.
- Increase deduplication threshold in SubGHz history from 500 to 600 ms.

-----
# For the reviewer

### DIFF
--- a/applications/main/subghz/subghz_history.c
+++ b/applications/main/subghz/subghz_history.c
@@ -265,7 +265,6 @@ bool subghz_history_add_to_history(
     instance->code_last_hash_data = hash_data;
     instance->last_update_timestamp = furi_get_tick();
 
-    FuriString* text = furi_string_alloc();
     SubGhzHistoryItem* item = SubGhzHistoryItemArray_push_raw(instance->history->data);
     item->preset = malloc(sizeof(SubGhzRadioPreset));
     item->type = decoder_base->protocol->type;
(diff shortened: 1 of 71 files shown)


WillyJL:
Adding shorthand's there would get cumbersome quickly with how many protocols there are, and I don't think starting to do that would be the best idea.

We can use a shorthand in the protocols name, could do BiSec or BSec or BS instead of the full "BiSecur"

user890104:
What about adding a new API method, so the protocol can provide code to generate its SubGHz history line string? I see you already added a custom member to the struct (long hash), so the BC is already broken.

For this protocol there's lots of data, so in order to make sense I would format is as follows:
- "HBS Wrong C/S" if the checksum does not match
- "HBS XX/YY/ZZ" if the checksum is correct where XX is the type (0x50 or 0x70), YY is a XOR hash of the serial number of the remote and ZZ is a XOR hash of the encrypted data. So each remote and each transmission from the same remote can be differentiated by just looking at the subghz history list entries.

Unrelated to this, the three repetitions of the packet (generated with a single button press) are send 504.3 ms apart, so they appear as a separare decoded messages, since the duplocation threshold is set to 500 ms. Not sure if it makes sense to bump this, because the x2 and x3 markers at the start of the lines eat up valuable space.

user890104:
The final format, which looks good on the flipper screen is:
`HBS Bad checksum` if the checksum is incorrect
or
`HBS XXXXXXXX:YY` where `XXXXXXXX` is the serial number and `YY` is a hash of the encrypted bytes.

user890104:
A two-byte data hash will make things more clear, and still fit on the same line.

WillyJL (approved):
LGTM, thanks!

PR #137, opened 2024-05-26
NFC: Add parser for CSC Service Works Reloadable Cash Card

zinongli:
# What's new

- A parser plugin that can parse CSC Service Works cards. This type of card uses Mifare Classic 1k to store balance-related data in plain text, allowing for reading. 
- I used keys I got from nonces from washers in my building. The keys are then used for verification and reading in this parser.  This could limit its potential to read card of the same type that uses a different key. But also prevents false positives.
- I have no prior experience in C but I am down to spend more time on polishing this. Currently this parser works very slowly and I suspect it's due to the lack of optimization on my part. Any suggestions would be appreciated. 
- I can't seem to upload the dumps from both of my cards here for the reviewers to test on their own. Please let me know what would be the best way I can get that to you.
- The wordings I chose like "Refill Times" can sound unobvious because this is not my first language. If you have better word-choice ideas please let me know. 

If it's a normal card:

Screenshot-20240526-023832

Screenshot-20240526-023844

If it's a new card. The card price is read from the card itself. If there were no signature of refill from a vending machine, then the last refill value in the card would reflect the price for getting this card.

Screenshot-20240526-025514

-----
# For the reviewer


WillyJL:
thank you! ill get to reviewing the code shortly

> I can't seem to upload the dumps from both of my cards here for the reviewers to test on their own. Please let me know what would be the best way I can get that to you.

you should be able to drag and drop into the github text box. if not, open up a ticket on discord :D (i think you already have one, if so let us know there and we'll open a new one)

> The wordings I chose like "Refill Times" can sound unobvious because this is not my first language. If you have better word-choice ideas please let me know.

i assume "refill times" means how many times the card balance was refilled? and if it was never refilled, it shows "new card"? if so, maybe "Topped up X times" or "Top-up Count: X" could work

on the other hand, im not sure what the difference between "balance" and "refilled balance" is, and what "card usage left" means

zinongli:
Sure. I will send that to you on Discord. 

I see. Top-up count does sound better. Balance is the remaining balance while refilled balance is how much there were when last refilled. So for example if I refilled 20 dollars and spent 10, then I have balance of 10 and refilled balance of 20. Card usage left is a counter that I found on the card memory. It decrements by one each time I use it.

WillyJL:
nice, so maybe "Top-up Count", "Last Top-up", "Card Usages Left", and "Card Value" for new card, those sound good to me )

gonna get to reviewing properly tomorrow, and thanks for the files. also, the section with the checkboxes in the PR template is for the reviewer checking your code :D leave them unchecked, we will check them once we try the code

zinongli:
Cool! I just changed the wordings accordingly. Thanks and sorry about the checkboxes. This is my first time being involved in a public repo. Still have a lot to learn.

applications/main/nfc/plugins/supported_cards/csc.c
```diff
@@ -0,0 +1,203 @@
...
+    {.a = 0xF329F7AAEDFF, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xacffffffffff, .b = 0xffffffffffff},
```
WillyJL: is this 0xacffffffffff one intentional?

applications/main/nfc/plugins/supported_cards/csc.c
```diff
@@ -0,0 +1,203 @@
...
+
+static const MfClassicKeyPair csc_1k_keys[] = {
+    {.a = 0xF329F7AAEDFF, .b = 0xffffffffffff},
+    {.a = 0xF329F7AAEDFF, .b = 0xffffffffffff},
+    {.a = 0xF329F7AAEDFF, .b = 0xffffffffffff},
+    {.a = 0xF329F7AAEDFF, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
+    {.a = 0xffffffffffff, .b = 0xffffffffffff},
```
WillyJL: repetition should be fine to remove, also are the 0xffffffffffff real keys being used?

applications/main/nfc/plugins/supported_cards/csc.c
```diff
@@ -0,0 +1,203 @@
...
+    do {
+        // Verify key
+        const uint8_t ticket_sector_number = 1;
+
+        const MfClassicSectorTrailer* sec_tr =
+            mf_classic_get_sector_trailer_by_sector(data, ticket_sector_number);
+
+        const uint64_t key =
+            bit_lib_bytes_to_num_be(sec_tr->key_a.data, COUNT_OF(sec_tr->key_a.data));
+        if(key != csc_1k_keys[ticket_sector_number].a) break;
```
WillyJL: should check key B too if applicable

WillyJL (changes requested):
code looks good to me, only need some improvements on key verification, check the individual comments. i can add the key b verification if you want, but need more info from you on the value(s) of key b. also if possible check if key is same for others in same building? having only 1 key could be limiting if its not universal

WillyJL:
undraft when ready

applications/main/nfc/plugins/supported_cards/csc.c
zinongli: Nice catch. It's a typo. Thanks for pointing it out.

zinongli:
I completely agree about hardcoding key values affecting universal applicability. I have been learning more about the parser verification process and about how to avoid false positives & misses. From my understanding, inside the _verify function one can't access card memory. But the most robust way to verify is checking the format of the memory because they are ususally the most custom designed for the specific company. So I will move all verification steps into the _parser function and leave the task of decoding keys to the users. It's a MFC 1k anyway. 

To answer your question about Key B, I don't have the value of Key B but Key A suffices reading 16/16 blocks. I tried flipper nested but it didn't support this kind of card. And I don't have a pm3 at hand so... It will be a mystery to be solved. 

I will send in my commit shortly. Thanks for the review!

octopus8321:
I am not a developer but I greatly admire the work that all developers do. It's a great job they do

octopus8321:
I am not a developer but I greatly admire the work that all developers do. It's a great job they do

Jimmy23222 (approved):

WillyJL:
@zinongli sorry for the delay, i finally looked better at the implementation and it seems meaningful. i cleaned it up a little, remvoed read and verify rather than returning true. can you please verify if it still works correctly? thanks :D

WillyJL (approved):
LGTM, just waiting for confirmation of functionality

zinongli:
Works great on my end too. Ready for merge!

WillyJL:
thank you!

PR #143, opened 2024-06-09
JS: Add ADC (analog voltage) support to gpio library

jamisonderek:
# What's new

- ADC (Analog to Digital Conversion) is now available in the firmware.  This change exposes the functionality to JavaScript.  You can run the program without any wires and should see some voltage levels.  If you connect 0-2 Volt output to pins 2,3,4 (A7, A6, A4) the program will display the voltage in millivolts.  You can also just attach a wire to A7 and touch it to see some additional voltage displayed.

-----
# For the reviewer

### DIFF
--- a/applications/system/js_app/examples/apps/Scripts/Examples/adc.js
+++ b/applications/system/js_app/examples/apps/Scripts/Examples/adc.js
@@ -1,0 +1,42 @@
+// This is an example of how to use the analog pins (ADC) on the Flipper Zero. 
+// The example uses a reference voltage of 2048mV (2.048V), but you can also use 2500mV (2.5V). 
+// The example reads the values of the analog pins A7, A6, and A4 and prints them to the console. 
+// The example also checks if the value of A7 is twice the value of A6 and breaks the loop if it is. 
+// The example uses the analog pins A7, A6, and A4, but you can also use PC3, PC1, and PC0. 
+
+let gpio = require("gpio");
+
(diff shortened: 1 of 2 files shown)


jamisonderek:
I just realized that people new to the API might call readAnalog("PA7") without first calling startAnalog(), which would assert on NULL handle.  This update will return an error message instead.  I'm not sure if the type of error should be `MJS_SYNTAX_ERROR` or something else?

jamisonderek:
NOTE: Don't directly touch the input wires to 3V3 as that will exceed 2.0 volts reference voltage.

An easy way to test the script is to put wires in pin A7 [pin2], A6 [pin 3], and A4 [pin 4], GND [pin 8] and 3V3 [pin 9].  Run the script, each pin should be reading about 100mV.  Touch one of the input wires (A7, A6 or A4) and the data should change.  With another finger touch ground and the value should decrease.  Remove your finger from ground and instead touch one of the input wires and 3V3 and the data should increase.  Do not exceed 2.0 volts, but if you do, use GND and the pin to help pull it back to normal levels (you can even touch the pin directly to GND).

WillyJL:
hi! thanks for the PR :D

just FYI, the checkboxes in "For the reviewer" section are for us to keep track of testing on the PR, please leave them blank when submitting

looks pretty good! i made some minor changes to how handle state is managed, it mirrors what badusb, usbdisk, and serial do: error if you try to start when already started, or stop when not started. also used INTERNAL_ERROR, since those follow that pattern aswell

WillyJL (approved):

PR #152, opened 2024-07-02
Updater: Gzip resources dejavu, refactor for CompressStreamDecoder API

WillyJL:
# What's new

- Re-implement gzip resources compression after OFW refactor with heatshrink
- Why? heatshrink: 11mb, gzip: 8.9mb

-----
# For the reviewer


WillyJL:
previous implementation for reference: 9b41168e8517e62423f7ce87580abe32e538472c

not ready yet, just scaffolding in place now, need to implement gzip mode in compress lib and tar stream

PR #156, opened 2024-07-07
Infrared: Correct Samsung.ir CH+ command

skyhawkillusions:
# What's new

- CH+ and CH- commands have same command currently
- CH+ is supposed to be 12 00 00 00 instead of 10 00 00 00
- Verified by comparing other files on UberGuidoZ repo and by checking own remotes

-----
# For the reviewer


WillyJL (approved):
thanks!

PR #159, opened 2024-07-10
Infrared: Added missing Koro Box entry to TV universal remote

skyhawkillusions:
# What's new

This is a combination of multiple changes that modified the databases to keep them up to date:
- tv.ir has been updated ~~based on a version of the file I found on an RM SD card~~ with Koro Box entry which is in OFW's tv.ir
- ~~audio.ir has been updated based on a version of the file I found on an RM SD card~~
- ~~NFC key dictionary has been updated based on the latest RRG file~~

-----
# For the reviewer


WillyJL:
we talked on discord, i will summarize here for transaprency. we have our mfc dict organized in a way that simplifies updating, i have gone ahead and updated it myself, thanks for the reminder to do so.

as for IR assets, most seemed to be duplicates (hence why i had already removed them in the past when merging from upstream), but Koro box is not. seems like this is all that remains from this PR.

regardless, thank you and i appreciate the effort :D

WillyJL (approved):

PR #162, opened 2024-07-16
Changing icons to make all backgrounds consistent

skyhawkillusions:
# What's new

I went through the files in the icons and changed them all to have a white background, where they were either transparent or orange, for consistency
-----
# For the reviewer


WillyJL:
@skyhawkillusions thanks for the PR! i would rather you submit this kind of change to official firmware repo, since it brings no tangible advantage after compilation, and since the files differ from official firmware ones it will cause merge conflicts and confusion if things change upstream (dealing with these is of course possible, im just saying that dealing with them is more annoying than having some icons orange or transaprent - if instead upstream merges this, then we are all on the same page)

skyhawkillusions:
> @skyhawkillusions thanks for the PR! i would rather you submit this kind of change to official firmware repo, since it brings no tangible advantage after compilation, and since the files differ from official firmware ones it will cause merge conflicts and confusion if things change upstream (dealing with these is of course possible, im just saying that dealing with them is more annoying than having some icons orange or transaprent - if instead upstream merges this, then we are all on the same page)

https://github.com/flipperdevices/flipperzero-firmware/pull/3802 The devs are currently automating the conversion to 1-bit images

WillyJL:
since this is being handled in OFW build process, i will close the PR here. still, thank you!

PR #172, opened 2024-07-30
docs: update ReadMe.md

eltociear:
# What's new

- movevent -> movement

-----
# For the reviewer


WillyJL (approved):
good catch )

PR #183, opened 2024-08-14
feat(subghz): added Solight TE44 recognition

m7i-org:
# What's new

Added the Solight TE44 weather station, with a protocol similar to Auriol HG06061A but with a checksum and different sync values. 
Updated the tests as well, including a raw and a decoded capture.

-----
# For the reviewer

- [x] I've uploaded the firmware with this patch to a device and verified its functionality (both manual and unit tests)


WillyJL (approved):
LGTM! thanks :D

m7i-org:
Thanks for fixing the linting, I couldn't get clang-format to work, somehow it didn't recognize some entries in `.clang-format`...

```
$ clang-format -i lib/subghz/protocols/solight_te44.h
YAML:28:3: error: unexpected scalar
  Enabled:         true
  ^
YAML:28:3: error: invalid boolean
  Enabled:         true
  ^
Error reading ./.clang-format: Invalid argument
```

WillyJL:
Might possibly be an older version that you have installed. Flipper includes it in the toolchain, you can do `./fbt format` and it should run fine :D

There's also `format_py` and `format_img`, or `format_all`

PR #187, opened 2024-08-21
Added backdoor keys

goosvorbook:
# What's new

- Added backdoor keys from source: https://eprint.iacr.org/2024/1275.pdf

-----
# For the reviewer


goosvorbook:
Nevermind, it seems like this requires some additional work https://news.ycombinator.com/item?id=41270384

WillyJL:
yep, checked with @equipter and these keys are to be used in a different kind of authentication, and thus would not work in a normal dictionary attack. @noproto is working on implementing the actual attack, please be patient :D

PR #194, opened 2024-08-26
Update keeloq.c

muermy:
# What's new

- [Describe changes here]

-----
# For the reviewer


WillyJL:
Mind explaining? This doesn't make any sense

WillyJL:
KeeLoq is a dynamic (rolling) code protocol. Why would you change this parameter? This probably breaks KeeLoq signal support

PR #201, opened 2024-09-02
NFC: Saflok parser Date Rollover

ghost:
Update the saflok parser to properly handle date rollovers.

# Saflok plugin date rollover

- The Saflok parser now handles dates rolling over properly. For example, a car programmed with a creation date of 8/31/2024 and an interval of 2 days would say 8/33/2024 instead of 9/02/2024. This is now been corrected to properly display expiring dates. 

-----
# For the reviewer


WillyJL (approved):
LGTM, thanks!

PR #203, opened 2024-09-02
SmartRider NFC Parser

jaylikesbunda:
## What's new
- Adds support for parsing SmartRider cards, the public transport smart card used in Western Australia.
- Implements reading and verification of SmartRider cards (Mifare Classic 1K).
- Displays key information:
  - Balance
  - Serial number
  - Concession type
  - Auto load threshold
  - Value amount
  - Information about last ten transactions

## Testing
- Tested with multiple SmartRider cards of varying types and usage patterns
- A sample card dump for testing is provided in the file: `Smartrider.nfc.txt`

## Checklist for reviewers
- [x] Verify correct parsing of card data
- [x] Confirm accurate display of all key information
- [x] Check for Mi-Fare Classic 1K false positives
- [x] Uploaded the firmware with this patch to a device and verified its functionality
- [x] Review error handling and edge cases

## Additional Notes

used data from https://github.com/SkryptKiddie/smartrider


ZProLegend007 (approved):
Exciting!

WillyJL (changes requested):
need to add more checks than just "the blocks we need have been read", currently it gives false positives for any mifare classic 1k.
could check if keys match a known set of keys, or could check the layout and content of data.
important is that parse() returns true only if this card is actually a SmartRider card

jaylikesbunda:
unfortunately i don't have other mifare classic cards to test with but i can confirm it reads and compiles correctly still with actual verification checks

WillyJL:
the checks are required in parse() too, this is what is giving false positives

jaylikesbunda:
sorry for the runaround, hopefully it's working correctly now. can confirm still builds correctly

jaylikesbunda:
recieved some cuid cards today so was able to test for myself can confirm works... **finally**
sorry bout the runaround, let me know if anything else needs changing

RogueMaster:
Hi @jaylikesbunda , seems like you have an extra file here. Please correct the right version of the file in the right location and re-test it is working as intended.

jaylikesbunda:
i can't believe i missed that thank you.

jaylikesbunda:
> Hi @jaylikesbunda , seems like you have an extra file here. Please correct the right version of the file in the right location and re-test it is working as intended.

i can't believe i missed that thank you. everything is still working just uploaded the formatting commit incorrectly

**_if i make any more commits after this point they will be fully tested and just because im bored_**

WillyJL (approved):
LGTM, nice work :D

ZProLegend007:
I updated to the latest commit and tested on my own smartrider, plus another to double check and it didn't read any different than it used to. Just tries and fails to unlock with classic dictionary and presents it as a plain mifare card 😕🧐

Persistent after a reinstall

WillyJL:
because you did not grab the keys. you need the keys. you can use nested and mfkey to crack them most likely. or install #207 which has noproto's mifare classic improvements, and managed to get all keys on its own

jaylikesbunda:
> I updated to the latest commit and tested on my own smartrider, plus another to double check and it didn't read any different than it used to. Just tries and fails to unlock with classic dictionary and presents it as a plain mifare card 😕🧐
> 
> Persistent after a reinstall

what willy said, you need to be able to read the whole card so you gotta crack the keys and make sure it says
Keys Found 32/32
Sectors Read 16/16
maybe someday will be able to figure out how the unique keys are generated so can unlock on the fly

WillyJL:
KDF (key derivation function) is the term :D yes would be cool, but with noproto's MFC improvements on the way it might not be needed after all xD

ZProLegend007:
Ahh I see, thank you very much for the elaboration. 👍

PR #212, opened 2024-09-11
NFC: Better mifare classic dictionary

wutchdugsinrl:
# What's new

- [Describe changes here]

-----
# For the reviewer


ZProLegend007:
Woah... awesome- but uhh... does the flipper have hardware capable enough to use all that information?? It's likely that it would take the flipper waaaaayyyy too long. Perhaps sorting the keys from most likely to be needed to least likely would help but even then that's wayyyy to large a task.

wutchdugsinrl:
hi yes I've tried it and the dictionary had all the keys for all my card

Drjacky:
@Willy-JL I see this PR got merged but why the `dev` still has the [old (small) dict](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/applications/main/nfc/resources/nfc/assets/mf_classic_dict.nfc)?

WillyJL:
@Drjacky we were all hopeful that it would be abundantly clear without further comment that this PR is a joke. A million lines of dictionary attack are useless and stupid, this was done as an inside joke making fun of one particular fake flipper influencer maliciously instructing people on how to waste time and get in legal trouble with huge dictionaries and fuzzing files like this one. I merged it as a joke and then immediately force pushed the branch back to the previous commit.

PR #215, opened 2024-09-13
Show satellites with an icon

m7i-org:
Replaces the display of the number of satellites during SubGHz scanning. If GPS is enabled (an an external module connected) it will blink a GPS icon with the number of sats currently in range (even if zero). Using a 10x10px "positioning" icon that replaces the SubGHz icon on the bottom right.

# What's new

Screenshot:
preview

- New "positioning" icon added, API bump to 3.1
  - Couldn't draw a nice 10x10px satellite, it's too small unfortunately. Hope it still makes sense though.
- Switches once per second from showing the number of items in the history vs the satellite number (even if 0) with an icon
- Replaces the previous satellite number as "X sats" as it never fit on the screen anyway

-----
# For the reviewer


WillyJL:
neat idea! maybe the pin could be 1 pixel shorter, ideal would be a bit thinner towards the bottom but its already the thinnest it can be, so maybe 1 pixel shorter. ill see if it looks better

WillyJL (approved):
good stuff <3

PR #216, opened 2024-09-14
Update mf_classic_dict.nfc

z3r0l1nk:
# What's new

Added 6 new keys from Bulgaria Hotel
Keys are not duplicated!

-----
# For the reviewer


WillyJL (approved):
thanks!

PR #217, opened 2024-09-15
Add Bresser 3CH weather station

m7i-org:
# What's new

Added the Bresser 3CH, a weather station supported by rtl_433 already in a stable state.
 - Protocol contains temperature, humidity, id, battery status, button press and a checksum.
 - Sometimes rtl_433 mixes this up with lacrosse (which does not have a checksum) so putting bresser before it just in case.

-----
# For the reviewer


WillyJL (approved):
LGTM! nice work

PR #225, opened 2024-09-17
Add API to enforce ISO15693 mode

aaronjamt:
See https://github.com/flipperdevices/flipperzero-firmware/pull/3885

# What's new

- Adds a new API to the NFC parser to allow applications to enforce a specific parsing mode. Some apps, such as Picopass, will always use one of the modes, so disabling autodetection can help to prevent issues resulting from incorrectly autodetecting the state (i.e. due to noise). It also includes a way to re-enable autodetection, should that be useful in the future.

This PR is identical to flipperdevices/flipperzero-firmware#3885.

-----
# For the reviewer


WillyJL (approved):
LGTM!

PR #237, opened 2024-09-27
Add support for NTAG I2C Plus 1k and 2k chips

RocketGod-git:
# What's new

Add support for NTAG I2C Plus 1k and 2k chips so my xSIID implant scans and displays NDEF properly.

-----
# For the reviewer


RocketGod-git:
Screenshot-20240926-172939

WillyJL:
interesting, i distinctly remember generating some I2C Plus tags and seeing a different memory layout, thats why i did not implement them as i was unfamiliar with them and did not have real hardware, and the ones i could generate looked different in structure. hey if it works, nice!

WillyJL (approved):

PR #244, opened 2024-10-05
Power: Suppress Shutdown on Idle While Charging / Plugged In

luu176:
This implements safeguard to prevent the flipper zero from shutting down while it's in a charging state. By doing so, we ensure that the battery charges fully without interruption.

# What's new

- Modified the `power_auto_shutdown_timer_callback` function to check the power state.
- If the device is charging, the shutdown process is suppressed, and the original shutdown timer is restarted based on the configured `shutdown_idle_delay_ms`
- The `shutdown_idle_delay_ms` value is now dynamically retrieved from the PowerSettings to ensure the timer is set accurately according to user-defined settings.


applications/services/power/power_service/power.c
```diff
@@ -333,8 +335,19 @@ static void power_storage_callback(const void* message, void* context) {
 static void power_auto_shutdown_timer_callback(void* context) {
     furi_assert(context);
     Power* power = context;
-    power_auto_shutdown_inhibit(power);
-    power_off(power);
+
+    // retreive shutdown delay from settings (in ms)
+    PowerSettings settings; 
+    power_settings_load(&settings); 
+    uint32_t shutdown_delay = settings.shutdown_idle_delay_ms;
```
WillyJL: this value is already accessible from `Power` service struct
```suggestion
```

applications/services/power/power_service/power.c
```diff
@@ -333,8 +335,19 @@ static void power_storage_callback(const void* message, void* context) {
...
-    power_off(power);
+
+    // retreive shutdown delay from settings (in ms)
+    PowerSettings settings; 
+    power_settings_load(&settings); 
+    uint32_t shutdown_delay = settings.shutdown_idle_delay_ms; 
+
+    // suppress shutdown on idle while charging to avoid the battery from not charging fully. Then restart timer back to original timeout.
+    if (power->state == PowerStateCharging) { 
+        furi_timer_restart(power->auto_shutdown_timer, shutdown_delay);
```
WillyJL:
```suggestion
        power_auto_shutdown_arm(power);
```

WillyJL (changes requested):
also please run `./fbt format`

WillyJL:
> The `shutdown_idle_delay_ms` value is now dynamically retrieved from the PowerSettings to ensure the timer is set accurately according to user-defined settings.

no point, when user changes this value, it is updated in service accordingly. this is just loading from file when theres no need to :D

WillyJL (approved):

PR #246, opened 2024-10-06
Sub-GHz: Fix GPS "Latitute" typo, switch to "Lat" and "Lon" in .sub files

m7i-org:
# What's new

- Fixed a typoin the SubGHz GPS coordinates for reading/writing sub files

-----
# For the reviewer


WillyJL (changes requested):
this is a breaking change, existing files would not be parsed correctly after this change. if you want, you can add backwards support for the typo version

WillyJL:
also there are others that rely on the broken spelling, since it has been there for so long. for example, hackrf portapack mayhem firmware has a wardrive map which supports flipper sub files with latitute typo. @htotoo might want to keep this in mind if we do go ahead

htotoo:
Thanks for the ping, will check on it!

m7i-org:
Fair point, it is indeed a breaking change. 

Shall we then rather agree on keeping `"Latitute","Longitude"` retro-compatible for reading sub files and moving on to `"Lat","Lon"` for reading/writing from now on? 

Then I think the switch makes more sense.

m7i-org:
Switched to "Lat","Lon" and looking for the older values in case they're not found. 

Added an extra check to make sure that we don't run in the [null island](https://en.wikipedia.org/wiki/Null_Island) problem when reading sub-files.

htotoo:
Imho lat / lon seems a good solution also the fallback part. I'll implement it the same way in Portapack.

WillyJL (approved):
LGTM :100:

PR #255, opened 2024-10-10
Infrared: Add option to "Load from Library File" for Universal Remotes

zxkmm:
# What's new

- add an entrance in IR universal submenu to load custom db

-----
# For the reviewer


applications/main/infrared/infrared_brute_force.c
```diff
@@ -60,32 +62,34 @@ InfraredErrorCode infrared_brute_force_calculate_messages(InfraredBruteForce* br
...
     InfraredSignal* signal = infrared_signal_alloc();
 
-    do {
-        if(!flipper_format_buffered_file_open_existing(ff, brute_force->db_filename)) {
-            error = InfraredErrorCodeFileOperationFailed;
-            break;
-        }
-
-        bool signals_valid = false;
+    if(!flipper_format_buffered_file_open_existing(ff, brute_force->db_filename)) {
```
zxkmm: replaced do-while loop to make sure it go through entire of  the db, performance shouldn't be impacted.

applications/main/infrared/infrared_brute_force.c
```diff
@@ -160,10 +170,40 @@ void infrared_brute_force_add_record(
...
+
     furi_string_free(key);
 }
 
 void infrared_brute_force_reset(InfraredBruteForce* brute_force) {
     furi_assert(!brute_force->is_started);
     InfraredBruteForceRecordDict_reset(brute_force->records);
 }
+
+size_t infrared_brute_force_get_db_size(const InfraredBruteForce* brute_force) {
```
zxkmm: this is needed to add btns on run time. i think ti won't make much conflicts when merging from upstream.

applications/main/infrared/scenes/infrared_scene_universal_more_devices.c
```diff
@@ -0,0 +1,178 @@
...
+    InfraredApp* infrared = context;
+    const InfraredErrorCode error = infrared_brute_force_calculate_messages(infrared->brute_force);
+    view_dispatcher_send_custom_event(
+        infrared->view_dispatcher,
+        infrared_custom_event_pack(InfraredCustomEventTypeTaskFinished, 0));
+
+    return error;
+}
+
+void infrared_scene_universal_more_devices_on_enter(void* context) {
```
zxkmm: in this func, note that it's actually loaded the db twice, here's what i did and why it's harmless:
1. load db previously cuz need to use the db to add btns in runtime
2. load db again with blocking task
reason:
1. there's a funny policy in infrared_brute_force_calculate_messages func, that it'll check if name in it, if yes, it will increase the num and add, if yes, it will make a new field.
it's probably a work around to satisfying other places, but since momentum is not a distro, i really don't want to edit that func to make it's hard to merge upstream changes.
why it's harmless: 
1. do it twice can make it cover the first custom item.
2. only the count (which controls the for loop to go through the next item by name field) x2, not index, so signal index not impacted.

also added this into comment to prevent misleading

applications/main/infrared/views/infrared_progress_view.c
```diff
@@ -81,9 +81,9 @@ bool infrared_progress_view_input_callback(InputEvent* event, void* context) {
         if(instance->back_callback) {
             instance->back_callback(instance->context);
         }
+        return true;
```
zxkmm: this is a work around because button_menu locked the operation, and i can only either:
1. make an un-lock func (this would change the API version so would be better to do on upstream)
2. make this don't pass even to button menu so it won't lock the button operation.

i choosed the second one

applications/main/infrared/scenes/infrared_scene_universal.c
```diff
@@ -84,6 +85,13 @@ void infrared_scene_universal_on_enter(void* context) {
         infrared_scene_universal_submenu_callback,
         context);
 
+    submenu_add_item(
+        submenu,
+        "More Devices",
```
jaylikesbunda: rename to "Custom DB" for easier understanding of the function?

zxkmm: Sure

applications/main/infrared/infrared_brute_force.c
WillyJL: it seems like the previous code already read the whole DB, only breaking if a signal is invalid, which this new version also does. isnt the only difference that now you also add the button record and keep track of index?

WillyJL: also loading records shouldnt change behavior for known remotes i think, perhaps we can add an internal flag or parameter to say if we want to load unknown buttons or not. this isnt exposed to api so it should be fine

applications/main/infrared/infrared_brute_force.c
WillyJL: get_button_count() ?

applications/main/infrared/scenes/infrared_scene_universal.c
```diff
@@ -84,6 +85,13 @@ void infrared_scene_universal_on_enter(void* context) {
         infrared_scene_universal_submenu_callback,
         context);
 
+    submenu_add_item(
+        submenu,
+        "Load Custom DB",
```
WillyJL: i know you just changed it but i think adding "from file" in some way would make this clearer. maybe "Load DB From File" ? also other references of "more devices" in enums and scene names can be changed for code clarity

applications/main/infrared/scenes/infrared_scene_universal_more_devices.c
```diff
@@ -0,0 +1,178 @@
...
+
+        view_set_orientation(view_stack_get_view(infrared->view_stack), ViewOrientationVertical);
+        view_stack_add_view(infrared->view_stack, button_menu_get_view(infrared->button_menu));
+
+        // Load universal remote data in background
+        infrared_blocking_task_start(
+            infrared, infrared_scene_universal_more_devices_task_callback);
+        view_dispatcher_switch_to_view(infrared->view_dispatcher, InfraredViewStack);
+    } else {
+        scene_manager_previous_scene(infrared->scene_manager);
```
WillyJL: else after long if block can be changed to a "guard check", change check to if(!condition), add previous scene there and return from this func

applications/main/infrared/scenes/infrared_scene_universal_more_devices.c
WillyJL: looking at how the blocking task works i think this can be done without loading the db twice, ill play around with it a bit

WillyJL (changes requested):
pretty solid work, but a few things can be ironed out. ill work on these improvements right away :smile:

applications/main/infrared/scenes/infrared_scene_universal_more_devices.c
WillyJL: hopefully, instead of loading it in on_enter, we can do the "initial" load in the blocking task callback and add the buttons from there. as for the first button detected not counting to total count, can just edit the logic in calculate_messages() since this is extra functionality added here anyway

applications/main/infrared/views/infrared_progress_view.c
WillyJL: i didnt understand what you meant here at first, but i see it now. ButtonMenu uses `freeze_input` to stop directional inputs while you're holding down a button, but in our case after pressing ok the progress view is shown so ButtonMenu does not receive the release event since the progress view consumes all inputs. and so ButtonMenu remains frozen.

rather than passing all inputs and only consuming short back press events, we could add a case before where if the event involves ok button and release type, it is not consumed, and add a comment explaining this.

or even better, we could make ButtonMenu trigger the click action when Ok is released, rather than when a short press is triggered. then its a small internal behavior change, and API is not impacted :D

WillyJL (approved):
i ended up changing more than i intended, but i think its now in great shape :D what do you think @zxkmm ?

zxkmm:
Ohhhhh, your change is so nice and elegant! Thank you so much! I didn't realize I missed that many parts! It's really nice and LGTM.

Also regarding the changelog name I'm not sure if i deserver it, since you re-wrote the entire work (with better code) already, so maybe add your name too?

Thank you so much and it's working better now

PR #256, opened 2024-10-10
Sub-GHz: Nexus-TH weather station protocol improvements on detection

m7i-org:
# What's new

Added detection improvements for Nexus-TH as in https://github.com/merbanan/rtl_433/blob/00bd97c63a58083d2a79ccb7089786ff9d0fda2e/src/devices/nexus.c#L63-L78 :
 - If hum==0% there's no humidity sensor, so don't overwrite to 20%
 - If the Rubicson check matches, then it's probably not a Nexus-TH
 - Check temp range beween -50C and 100C

I haven't found the time of looking into changing the display name (rtl_433 calls the sensor a Nexus-T if there's no humidity), but I think this is good enough.

-----
# For the reviewer


m7i-org:
@htotoo your .sub-generator is linked in the source, I imagine you might have something to say about the changes?

WillyJL (approved):

PR #258, opened 2024-10-12
GUI: Add a fast-select "Zapper" mode for VarItemList

zxkmm:
# What's new

Add a fast-select "Zapper" function for var_item type menu

- in varible_item menu (e.g. the freq select menu), in old way you can only use Left and Right to swap.
- I added a "Zapper menu", you highlight a item, long press
- then use OK to paging, and use ^ v < > to pick one, kinda likes cheat code in old games
- benefits:
  it's 4 times faster when you are looking for a item (e.g. a frequency)
  I't s extremely fast (and don't even need to look at the screen) if you remember the number (position) of that specific item (for example the most used two: 315 and 433.92)
```
go to 433.92: ok*9, then Right
go to 315.00: ok*4, then Right
```

~~TODO~~

~~Current interface is kinda ugly. Need to design a new one maybe.~~

-----
# For the reviewer


applications/services/gui/modules/variable_item_list.c
```diff
@@ -220,6 +328,65 @@ void variable_item_list_set_header(VariableItemList* variable_item_list, const c
         true);
 }
 
+static bool zapper_menu_input_handler(InputEvent* event, void* context) {
```
zxkmm, line 323: in this (and draw) func, if it changes real freq from cc1101 part, then probably need a real buffer, instead of doing "change 4 times and paint", to prevent flood to 1101 and cause err-ish. 

but from my read, it seems like it only save - not changing, unless back-button as call back. so should be OK and saved more RAM. but need check more (maybe)

applications/services/gui/modules/variable_item_list.c
```diff
@@ -229,9 +396,17 @@ static bool variable_item_list_input_callback(InputEvent* event, void* context)
...
+        {
+            if(model->is_zapper_menu_active) {
+                consumed = zapper_menu_input_handler(event, context);
+            } else {
+                locked_message_visible = model->locked_message_visible;
+            }
+        },
         false);
 
+    if(consumed) return true;
```
zxkmm, line 421: it's for prevent the zapper menu infect in the var_item meny obj.

applications/services/gui/modules/variable_item_list.c
```diff
@@ -37,21 +45,121 @@ typedef struct {
...
 static void variable_item_list_process_ok(VariableItemList* variable_item_list);
+static void variable_item_list_process_ok_long(VariableItemList* variable_item_list);
 
 static size_t variable_item_list_items_on_screen(VariableItemListModel* model) {
     size_t res = 4;
     return (furi_string_empty(model->header)) ? res : res - 1;
 }
 
+static void zapper_menu_draw(Canvas* canvas, VariableItemListModel* model) {
+    // TODO: ugly now
```
zxkmm: i'm bad at designing UI but feel free to draw a draft and i can implement it

applications/services/gui/modules/variable_item_list.c
```diff
@@ -285,6 +460,15 @@ static bool variable_item_list_input_callback(InputEvent* event, void* context)
         default:
             break;
         }
+    } else if(event->type == InputTypeLong) {
```
zxkmm, line 476: only one option but switch to keep it clean

applications/services/gui/modules/variable_item_list.c
```diff
@@ -413,6 +597,42 @@ void variable_item_list_process_ok(VariableItemList* variable_item_list) {
...
+void variable_item_list_process_ok_long(VariableItemList* variable_item_list) {
+    furi_check(variable_item_list);
+
+    with_view_model(
+        variable_item_list->view,
+        VariableItemListModel * model,
+        {
+            VariableItem* item = variable_item_list_get_selected_item(model);
+
+            bool is_allowed = false;
```
zxkmm, line 622: this is just for prevent mistaken-touch for those two options field. but they works good too, just i think not needed

applications/services/gui/modules/variable_item_list.c
zxkmm: just checked and it's not, so it's safe.

zxkmm:
~~Just broken changelog, lol, will fix now~~
it turned out is github's diff view is broken

PR #261, opened 2024-10-13
fix that load custom db always-1 comparing the read db

zxkmm:
# What's new

- in my original solution is call the calculate db twice to resolve this.
- but since Willy already added one more arg for the calculate func, then can easily use it as indicator to fix the issue.

-----
# For the reviewer


WillyJL:
are you sure this is even required anymore? i just tested on dev branch, loading from file or loading normally from preset gives same numbers: digital sign power has 5 signals and play has 5 signals, tv power has 296 signals. same results when loading custom db

zxkmm:
last time when I try test branch it’s still -1, lemme try dev build now

zxkmm:
ohhhh shiii it looks like fixed in the dev build. Maybe my previous build isn't clean or something, lol

PR #265, opened 2024-10-14
NFC: NDEF parser Mifare Classic support, protocol-agnostic refactor and more improvements

luu176:
# What's new

- Added Mifare Classic support
- Reads the MAD (Mifare Application Directory) and finds all sectors that possibly have ndef
- Once ndef data is found, parsing begins.

-----
# For the reviewer


WillyJL (approved):
i implemented the changes i talked to you about on discord, now ndef is addressed with a protocol independent position value in "data space", then each protocol translates this to data on the card as needed. for mfc, this means as you had done, to skip sector trailers. also, for mfc 4k, there is mad2 in sector 16 that gets skipped. i dont have any ndef mfc 4k files to test on, but it should index mad2 correctly too now.

i apologize, i changed so much that there isn't much recognizable from what you submitted, but this had to be done eventually since we dont have enough ram to spare on a cloned buffer. still, your code really helped me understand how ndef on mfc works, would have been a struggle without.

also changed other things like uri decoding and type name handling to avoid heap allocations at all costs.

PR #267, opened 2024-10-14
NFC: Improve URI decoding in NDEF parser

jaylikesbunda:
# What's new

- New decode_char function for hex to ASCII conversion
- Updated parse_ndef_uri with decoding step
- Increased buffer sizes to prevent overflows
- Added memory allocation checks

Screenshot-20241015-060628

-----
# For the reviewer


applications/main/nfc/plugins/supported_cards/ndef.c
```diff
@@ -86,29 +98,56 @@ static void parse_ndef_uri(FuriString* str, const uint8_t* payload, uint32_t pay
...
     size_t prepend_len = strlen(prepend);
-
     size_t uri_len = prepend_len + (payload_len - 1);
-    char* const uri_buf = malloc(uri_len);
+
+    char* uri_buf = malloc(uri_len + 1);
+    if(!uri_buf) {
+        furi_string_cat_printf(str, "Memory allocation failed\n");
+        return;
+    }
```
WillyJL: we are lucky, flipper malloc() will crash on OOM, this check can go

applications/main/nfc/plugins/supported_cards/ndef.c
```diff
@@ -86,29 +98,56 @@ static void parse_ndef_uri(FuriString* str, const uint8_t* payload, uint32_t pay
...
+
     memcpy(uri_buf, prepend, prepend_len);
     memcpy(uri_buf + prepend_len, payload + 1, payload_len - 1);
-    char* uri = uri_buf;
+    uri_buf[uri_len] = '\0';
```
WillyJL: and flipper malloc() also does memset(0) so its already \0 at the end

applications/main/nfc/plugins/supported_cards/ndef.c
```diff
@@ -86,29 +98,56 @@ static void parse_ndef_uri(FuriString* str, const uint8_t* payload, uint32_t pay
...
+
+    char* decoded_uri =
+        malloc(uri_len * 2 + 1); // Worst case scenario: every character is percent-encoded
+    if(!decoded_uri) {
+        furi_string_cat_printf(str, "Memory allocation failed\n");
+        free(uri_buf);
+        return;
+    }
```
WillyJL: same as above, no need to check failed malloc()

applications/main/nfc/plugins/supported_cards/ndef.c
```diff
@@ -86,29 +98,56 @@ static void parse_ndef_uri(FuriString* str, const uint8_t* payload, uint32_t pay
...
+
     memcpy(uri_buf, prepend, prepend_len);
     memcpy(uri_buf + prepend_len, payload + 1, payload_len - 1);
-    char* uri = uri_buf;
+    uri_buf[uri_len] = '\0';
+
+    char* decoded_uri =
+        malloc(uri_len * 2 + 1); // Worst case scenario: every character is percent-encoded
```
WillyJL: i think you got the right intent, but logic the wrong way around :D

if encoded its `"%20"`, 3 characters, decoded it becomes `" "`, so 1 character. so, worst case scenario, nothing was encoded, so length is the same. best case scenario, everything is encoded, length is roughly / 3. malloc() of same size should work fine?

WillyJL: and because of this, i think it is possible to decode without a second buffer... iterate over same string keeping 2 indexes (original and decoded), for each iteration check char at original index: 
if percent: replace char at decoded index with decoded char, increase decoded index by 1, increase original index by 3
if any other character: read char at original index and write to decoded index
because the indexes will always either increase at same pace, or original increases faster than decoded, you can overwrite data because original index is ahead
not sure if im making sense lol

applications/main/nfc/plugins/supported_cards/ndef.c
```diff
@@ -86,29 +98,56 @@ static void parse_ndef_uri(FuriString* str, const uint8_t* payload, uint32_t pay
...
+        furi_string_cat_printf(str, "Memory allocation failed\n");
+        free(uri_buf);
+        return;
+    }
+
+    size_t decoded_len = 0;
+    for(size_t i = 0; i < uri_len; i++) {
+        if(uri_buf[i] == '%' && i + 2 < uri_len) {
+            decoded_uri[decoded_len++] = decode_char(&uri_buf[i]);
+            i += 2;
```
WillyJL: this branch needs to check if next 2 characters are valid hex

applications/main/nfc/plugins/supported_cards/ndef.c
```diff
@@ -40,6 +40,18 @@ static void
...
+    int high = (str[1] >= 'A' && str[1] <= 'F') ? (str[1] - 'A' + 10) :
+               (str[1] >= 'a' && str[1] <= 'f') ? (str[1] - 'a' + 10) :
+               (str[1] >= '0' && str[1] <= '9') ? (str[1] - '0') :
+                                                  0;
+    int low = (str[2] >= 'A' && str[2] <= 'F') ? (str[2] - 'A' + 10) :
+              (str[2] >= 'a' && str[2] <= 'f') ? (str[2] - 'a' + 10) :
+              (str[2] >= '0' && str[2] <= '9') ? (str[2] - '0') :
+                                                 0;
+    return (high << 4) | low;
+}
```
WillyJL: i think there is a function like this in flipper libs, ill check

applications/main/nfc/plugins/supported_cards/ndef.c
WillyJL: but also depends if we want to show the original buffer...
fair, modern browsers will auto-encode if you type unsupported characters into the url bar, but still its not an accurate representation of what is on the card... eh should be fine

WillyJL (changes requested):
i can get those patched up :D

applications/main/nfc/plugins/supported_cards/ndef.c
jaylikesbunda: > but also depends if we want to show the original buffer... fair, modern browsers will auto-encode if you type unsupported characters into the url bar, but still its not an accurate representation of what is on the card... eh should be fine

personally i don't see the need if it's functioning correctly but yeah that's a really good point avoiding a second buffer

applications/main/nfc/plugins/supported_cards/ndef.c
WillyJL: not relevant anymore, i changed it to print exact number of chars, so dont even need a terminator, we know the length

applications/main/nfc/plugins/supported_cards/ndef.c
WillyJL: there would be sscanf() which would be cleaner code, but for this usecase it is overkill, uses much more stack

jaylikesbunda: would this work?

```
static inline int hex_to_int(char c) {
    if (c >= '0' && c <= '9') return c - '0';
    if (c >= 'A' && c <= 'F') return c - 'A' + 10;
    if (c >= 'a' && c <= 'f') return c - 'a' + 10;
    return 0;
}

static char decode_char(const char* str) {
    return (hex_to_int(str[1]) << 4) | hex_to_int(str[2]);
}
```

WillyJL (approved):

applications/main/nfc/plugins/supported_cards/ndef.c
WillyJL: i missed this oops. yeah that is a lot clearer code, nice!

PR #273, opened 2024-10-20
JS: Test moving JS runner into flash

WillyJL:
# What's new

- Might help with RAM issues with big scripts
- Need more flash, furi trace is already disabled so situation is bad

-----
# For the reviewer


WillyJL:
this was originally due to some users reporting out of memory crashes with scripts. @jamisonderek mentioned on discord that this might actually be due to the usbdisk module using a buffer too large for available ram. if it is possible to shrink this buffer, moving js_app back to dfu might not be necessary

WillyJL:
closing as stale, and also havent heard issues with js running out of ram as much lately

PR #276, opened 2024-10-24
CLI: More commands as plugins on SD, refactor plugin wrapper

WillyJL:
# What's new

- trying to free some flash space
- moving more cli commands to plugins on sd card
- cli command entry points are now more similar to OFW ones
- to make CLI plugin wrappers just need `CLI_PLUGIN_WRAPPER("name", callback)` macro and `"name_cli"` entry in app manifest
- removing unused icons from firmware blob
- fix duplicate subghz region loading, oops

-----
# For the reviewer


GameLord2011:
Great, now I'll need to re-write my custom commands!

PR #285, opened 2024-10-30
Update momentum_app.c

grugnoymeme:
# What's new

- [Describe changes here]

-----
# For the reviewer


WillyJL:
This is intentional... if you come from xtreme, this will make sure it gets converted to support momentum

PR #293, opened 2024-11-07
NFC: Merge OFW 3988: Fix ISO15693 stuck in wrong mode

WillyJL:
# What's new

- alternative fix to #225 

-----
# For the reviewer


WillyJL:
merged via f2011a446b78e35872c3bb35ba04d3361f1f9b3e

PR #310, opened 2024-11-24
NFC plugin: Ventra ULEV1 parser

hazardousvoltage:
# What's new

- Added support for Ventra ULEV1 to NFC parser.  This can parse the paper 'single-ride' and 'day pass' tickets, show the
pass expirations, number of rides remaining, where it was purchased and used, etc. in a concise and useful format.
- Does not handle the plastic DESFire Ventra.

-----
# For the reviewer


hazardousvoltage:
flipper-ventra-info

applications/main/nfc/plugins/supported_cards/ventra.c
```diff
@@ -0,0 +1,292 @@
...
+
+#include <flipper_application/flipper_application.h>
+#include <nfc/protocols/mf_ultralight/mf_ultralight.h>
+#include "datetime.h"
+#include <furi_hal.h>
+
+#define TAG "Ventra"
+
+DateTime ventra_exp_date = {0}, ventra_validity_date = {0};
+uint8_t ventra_high_seq = 0, ventra_cur_blk = 0, ventra_mins_active = 0;
```
WillyJL, line 24: i'd suggest moving these to parameters for the respective functions, but for the scope of this parser it should be fine like this too.

WillyJL (approved):
LGTM, and sorry for the delay. i'd like to test with an example .nfc file dump of a supported card if you can share one, meanwhile i'll merge as the code looks fine to me. thanks!

applications/main/nfc/plugins/supported_cards/ventra.c
hazardousvoltage: Yeah, the code could use a little cleanup indeed.. I finally found my rare 7-day Ventra card too, so I really should add that.  Probably the right thing to do is move all this into Metroflip, when I have the time.  Random dump attached for testing, it's 5 years old so I'm not too worried about violating anyone's privacy sharing it.

[MF_UL_11_2025-07-15_08,40,43.txt](https://github.com/user-attachments/files/21233832/MF_UL_11_2025-07-15_08.40.43.txt)

PR #314, opened 2024-12-01
Main Menu: Add coverflow menu style

CodyTolene:
# What's new

- Add support for scaling menu icons (0-100%)
- Add new menu style: "CoverFlow"

CoverFlow

-----
# For the reviewer


applications/services/gui/canvas.c
```diff
@@ -298,21 +298,37 @@ void canvas_draw_icon_animation(
     Canvas* canvas,
     int32_t x,
     int32_t y,
+    int32_t width_scale,
+    int32_t height_scale,
```
WillyJL, line 302: unfortunately we cannot change this, as this function is exposed to public api and must remain compatible across firmwares and also not break backwards compatibility. could make a separate function for this. to avoid code duplication, this current function could be made static and add a `_` in front, then make 2 new functions, one with original signature and one with scale arguments, that just call the common static function. the new function could be called ..._ex() to indicate the extended functionality

WillyJL (changes requested):
also please run `./fbt format`

applications/services/gui/canvas.c
CodyTolene: Good info thank you I will look into applying that now 👍

CodyTolene: Good to go. Thank you!

WillyJL:
i'll look closer (both at this and other open prs) and merge in the coming days, a bit swamped with uni deadlines right now. still, looks very cool, nice work!

CodyTolene:
Awesome sounds great @Willy-JL have a good night & happy holidays!

WillyJL (approved):
theres a few small tweaks i would make, just technicalities, but it doesnt let me push to your fork's branch, please enable "allow edits by maintainers" next time, makes it easier :D

regardless, overall looks good, thanks!

CodyTolene:
@Willy-JL Thanks man! Happy to contribute hopefully I can get some time for more in the future it sure was fun

956MB:
@CodyTolene I tried messaging you on Discord but perhaps you don't use it. Would you be open to a tiny change to your coverflow design like this? It refines the corners making them rounded (which is used a lot around the UI) and slightly fixes the spacing between the cards, text, scrollbar and surrounding border. It's no major redesign obviously, and keeps the spirit of your original design. It is your menu (which I like a lot), so it is up to you which is why I'm asking first.

#### Before

Original

#### After

Edits

CodyTolene:
This looks so good nice work on it, I really like it!

Thanks for pinging me @956MB I love seeing improvements like this. Also I haven't been on Discord in some time, I'll have to check that soon sorry about that. 

Looking forward to seeing your updates in future firmware! Have a good one

PR #322, opened 2024-12-15
Archive: Add dynamic paths to browser tab

956MB:
# What's new

Adds to the option to show current directory/path in archive tab instead of static tab text. Default state is `OFF` to maintain the existing look, but I personally think `Current` could probably be set as default if merged. Of course all text involed `Show Browser Path`, `Current`, `Brief` and `Full` could be changed as well.

- Adds setting under "Interface > File Browser > Show Browser Path" with four options:
  - Off (default): Shows static tab text like before (e.g. "Browser", "Sub-GHz", etc.)
  - Current: Shows the current directory name only (e.g. "Games")
  - Brief: Abbreviates the full path to the current directory (e.g. /e/a/Games")
  - Full: Shows complete path from root (e.g. "/ext/apps/Games")
- Path updates when navigating through directories. It is also maintained when opening an app and clicking back into the browser.
- Maintains existing header styling and layout keeping it within the frame
- Long paths that exceed the frame width are drawn with `elements_scrollable_text_line` to allow for full visibility.
    - e.g. Short paths like `Games` are drawn centered like normal, and long ones like `/ext/apps/Games` scroll.
    
#### "Off"

OFF

#### "Current"

Current

#### "Brief"

Brief

#### "Full"

Full

-----
# For the reviewer


WillyJL:
*missed opportunity for `/e/a/Sports` in the brief example xD*

sorry i just had to :laughing:, i like this idea! will review in the coming days

956MB:
> _missed opportunity for `/e/a/Sports` in the brief example xD_

AGH! Scrap the whole thing. Kidding, looking forward to it 👍

lib/momentum/settings.c
```diff
@@ -100,6 +101,7 @@ static const struct {
     {setting_bool(sort_dirs_first)},
     {setting_bool(show_hidden_files)},
     {setting_bool(show_internal_tab)},
+    {setting_bool(show_browser_path)},
```
WillyJL: using `setting_bool()` here will enforce the value is either 0 or 1, and save to file as "true" or "false". should be `setting_enum()`, which also specifies the max value (discards values that arent 0 to max-1). also since its a multiple option setting, could be renamed to `browser_path_mode` to match the enum.

applications/main/archive/views/archive_browser_view.c
```diff
@@ -293,7 +347,20 @@ static void archive_render_status_bar(Canvas* canvas, ArchiveBrowserViewModel* m
     canvas_draw_rframe(canvas, 0, 0, 51, 13, 1); // frame
     canvas_draw_line(canvas, 49, 1, 49, 11); // shadow right
     canvas_draw_line(canvas, 1, 11, 49, 11); // shadow bottom
-    canvas_draw_str_aligned(canvas, 25, 9, AlignCenter, AlignBottom, tab_name);
+
+    size_t text_width = canvas_string_width(canvas, tab_name);
```
956MB: I just realized looking at this that checking the width and using canvas_draw_str_aligned isnt necessary because scrollable_text_line will function the same when it doesn't need to scroll.

WillyJL: yep :D

applications/main/archive/views/archive_browser_view.c
```diff
@@ -58,6 +58,57 @@ void archive_browser_set_callback(
...
 }
 
+static void archive_update_formatted_path(ArchiveBrowserViewModel* model) {
+    ArchiveBrowserView* browser = model->archive->browser;
+    if(!browser->path_changed) {
+        return;
+    }
+
+    ArchiveTabEnum tab = archive_get_tab(browser);
+    if(!momentum_settings.show_browser_path || archive_is_home(browser)) {
```
WillyJL: here value 0 coincides with off, but being an enum (and also considering renaming the setting as before) its better to explicitly check against enum value

applications/main/archive/views/archive_browser_view.c
```diff
@@ -58,6 +58,57 @@ void archive_browser_set_callback(
...
+                }
+                furi_string_right(remaining, slash_pos + 1);
+            }
+
+            furi_string_free(token);
+            furi_string_free(remaining);
+            break;
+        }
+
+        default:
```
WillyJL: better to specify the enum value here too, in case there are more values added later, also makes it clearer

applications/main/archive/views/archive_browser_view.c
```diff
@@ -58,6 +58,57 @@ void archive_browser_set_callback(
...
+            path_extract_basename(path, browser->formatted_path);
+            break;
+        }
+    }
+
+    furi_string_set(browser->prev_path, browser->path);
```
WillyJL: this doesn't seem to be used anywhere, lost between refactoring? ;)

WillyJL (changes requested):
small things, working on them rn

applications/main/archive/views/archive_browser_view.c
WillyJL: actually no, if its shorter, then the scrollable text would align it to left instead of centered. however, there is a centered variant of the scrollable text function

WillyJL:
hmm tho the way i did it makes it not change text while searching. just that the scrollable text text functions need a furistring so i thought to just use the same one as formatted path, but then for searching it needs to change even if the path didnt change. might be easier to just use a temporary furistring, or to use the normal draw function while searching

WillyJL (approved):
nice :D

Issue #254, opened 2024-10-08
Archive: Allow selecting multiple files


AztecCodes:
### Describe the enhancement you're suggesting.

Add the function to the explorer that you can select multiple files and then make use of action like cut, delete and copy.
This would be great!

Anonymous3-a:
I feel like if you're doing anything more complex than one or two files, you should just use a computer to move the files, although that would be a nice feature.

Choder7:
Cool Suggestion! 👍

You would probably hold down OK in the file explorer and then click on SELECT, then you can just click on other files _without having to hold down OK and choose select again for every file._ When you selected every file you want to move, copy, ..., you just hold down and click on move, copy, ...

fiveangle:
This actually would be pretty cool !

956MB:
> You would probably hold down OK in the file explorer and then click on SELECT, then you can just click on other files _without having to hold down OK and choose select again for every file._ When you selected every file you want to move, copy, ..., you just hold down and click on move, copy, ...

Are we talking about file only selection, or files and folders? Because I think the process for selecting makes sense if you only want to do files or folders, not both at the same time. Is that the main goal with multiple selection here? Like one would only really be dealing with a handful of files in a folder at a time, rather than a large selection of files/folders in different locations? I haven't had the need/desire to move multiple things around yet, but I def think it *should* be a thing.

With the hold OK being for the Actions context menu, and the click OK being for selecting items in "select mode", how do you click in to select the files in some folder, and select an entire folder? Seems like its a potentially "keybind limited" action if you're trying to do more than files, right? Am I making this too complicated or just misunderstanding?

WillyJL:
i think having both file and folder selection would be ideal, as it reflects what you can do in desktop file managers for example. but you bring up a good point, that it has some extra concerns for inputs.

i had a couple ideas. all of them would start with holding ok, then entering select mode. then:
1. ~~press/hold ok on file to select, press ok on folder to enter it, hold ok on folder to select it~~ i realize after writing that this wouldnt work, as you wouldnt have a way to bring up the more actions menu as hold ok is already used
2. press left/right (not sure which, or if both) on file/folder to select, press ok on folder to enter it, hold ok on file/folder for actions menu, press ok on file does nothing (?)
3. press ok on file/folder to select, hold ok on file/folder for actions menu, press right on folder to enter it

[...]

PR #327, opened 2024-12-20
Archive: Allow selecting multiple files

956MB:
# What's new

Addresses #254 

As discussed in that issue, this implements multiple file/folder selection in the Archive file browser. With some personal design choices for the list selection indicators that can be changed if needed.

- Files and folders in different locations can be selected and then actions done on them (`Cut`, `Copy`, `Paste`, `Delete`)
- A check is used to make sure the destination path is not a child of one of the selected files/folders, and an error dialog is shown if tried.
- List items in the browser have either a regular indicator `[+]` (for files or entire folders), or a number `[2]` to show the amount of selected files in a folder. 
- The setting `Interface > File Browser > Selection Indicator` was added to change which character is used in the selected file indicator: `+`, `*` or `-`. (These can also be changed or more added.)
- When a folder has files selected inside it and shows a `[3]` indicator, the folder itself can be selected and all the files inside it are deselected and the folder shows a `[+]`.
- Renaming a file while in `select mode` is also possible, but it works on one at a time. Not all the selected files.

Keybinds for the "select mode" are as follows:
- *left*: deselect
- *right*: select
- *click OK*: enter folder
- *hold OK*: opens Actions/Manage menus
- *back*: exit folder

This set of keybinds allows for selecting files and folders the same way and the ability to enter folders for its contents. as mentioned in the [issue comments](https://github.com/Next-Flip/Momentum-Firmware/issues/254#issuecomment-2555186801).

Couple of other things I'd thought of doing but haven't yet (TODO):
1. Make the `Select` "badge" text at the top of the status bar also show the amount of selected files, like this `Select (4)`.
2. ~~Add a scrollable list to the delete screen when theres more than one file being deleted. Gives you a second to see the files about to be deleted rather than "Delete 5 files?".~~

#### Delete files in multiple folders

delete_files_in_multiple_folders

#### Copy folder and file

copy_folder_and_file

#### Copy files and delete

show_selected_files_and_delete

#### Rename selected file then delete

rename_selected_file_then_delete

#### Nested path error dialog

nested_path_error

-----

# For the reviewer


956MB:
I updated the delete screen with a new file list widget. It's scrollable up/down but the files are obviously not selectable with a cursor or anything. They're just there as a visual confirmation of what you're about to delete. Of course it's up to whoever in charge whether it should be in there of not, but I think it's helpful. Fills out the empty space too.


Screenshot 2024-12-29 at 10 51 26 AM

PR #331, opened 2024-12-25
Desktop / MNTM Settings: Directories and generic files support for Keybinds / Main Menu

956MB:
# What's new

Adds *RIGHT* button select in the file browser dialogs and changes the `Open File` action to `Open File/Directory` in `Settings > Desktop > Keybinds Setup`. This adds the ability to open to any directory in the Archive app, in addition to the default behavior of opening a file in it's default app.

ScreenRecording2024-12-25at6 19 41AM-ezgif com-video-to-gif-converter

-----
# For the reviewer


956MB:
The adding of JS files and other stuff to the main menu seemed connected to the initial PR with the `select_right` thing for the keybinds dialog, so I just added on to this PR instead of creating a new one for this feature.

### Screenshot of the files/folder with icons in the Main Menu

Screenshot-20241227-205546

Another thing I was thinking about but haven't messed with is whether the word `App` ought to still be used in the menu, as it can basically have any file type with this change, not just apps. For simplicity sake it'd probably be best just to keep it, though I think `File` can work well if you think of everything on the device as 'a file', including apps.

- "App 1/12" -> "File 1/12"
- "Add App" -> "Add File"
- "Move App" -> "Move FIle"
- "Remove App" -> "Remove File"

Just a thought.

WillyJL:
> The adding of JS files and other stuff to the main menu seemed connected to the initial PR with the `select_right` thing for the keybinds dialog, so I just added on to this PR instead of creating a new one for this feature.

very nice!



> Another thing I was thinking about but haven't messed with is whether the word `App` ought to still be used in the menu, as it can basically have any file type with this change, not just apps. For simplicity sake it'd probably be best just to keep it, though I think `File` can work well if you think of everything on the device as 'a file', including apps.
> 
> * "App 1/12" -> "File 1/12"
> * "Add App" -> "Add File"
> * "Move App" -> "Move FIle"
> * "Remove App" -> "Remove File"
> 
> Just a thought.

i think "Item" would work better as a single name for both apps and files. or alternatively, could use both app and file, changing the text of the buttons based on what is currently selected, but that is a bit over the top i think

956MB:
> i think "Item" would work better as a single name for both apps and files.

Oh yep, you're right. "Item" is generic enough for all of it 👍

applications/services/loader/loader_menu.c
```diff
@@ -153,7 +155,28 @@ static void loader_menu_apps_callback(void* context, uint32_t index) {
...
+        if(is_dir) {
+            Desktop* desktop = furi_record_open(RECORD_DESKTOP);
+            desktop->archive_dir = furi_string_alloc_set(menu_app->path);
+            furi_record_close(RECORD_DESKTOP);
+
+            furi_thread_flags_set(furi_thread_get_id(app->loader_menu->thread), 0);
+            if(app->loader_menu->closed_cb) app->loader_menu->closed_cb(app->loader_menu->context);
+
+            view_dispatcher_send_custom_event(
+                desktop->view_dispatcher, DesktopMainEventOpenArchive);
```
WillyJL: i get why its like this, but i hate it xD

ill try to come up with a way to launch archive without passing through desktop

WillyJL: reworked it into desktop with less FuriStrings being passed around, and also handled folders in run_with_default_app() for less duplicated code

WillyJL (approved):
*one eternity later*

WillyJL (approved):

956MB:
Merged :) #369

Issue #338, opened 2024-12-31
Battery percentage on MNTM menu style is off by one percent!


GameLord2011:
### Describe the bug.

The battery percent on the mntm menu style is one less then the percent shown on the desktop

### Reproduction

1. Turn on flipper zero.
2. Switch menu style to mntm.

### Target

Firmware, Hardware

### Anything else?

my suggestion for a fix is to change line 409 of `menu.c` to:
```C
snprintf(bat_display, sizeof(bat_display), "%d%%", battery_percent + 1);
```

956MB:
Probably better to use the function that's already there for that than to +1, I think. :)

956MB:
https://github.com/Next-Flip/Momentum-Firmware/compare/37d8bf76...dbbecd9e1fbf4449af56f1eec0c69979e7b8538f

PR #339, opened 2024-12-31
Main Menu: Fix MNTM style battery percent off by 1

956MB:
# What's new

#338 

- Uses the same `furi_hal_power_get_pct` function used on the desktop to fix the apparent rounding issue that's happening and show the same battery %.

#### Before

off-by-1-1

#### After

off-by-1-2

-----
# For the reviewer


GameLord2011:
Checked on personal fork, is stable!

WillyJL (approved):

Issue #317, opened 2024-12-08
12 Hour Clock Formatting Issues


GonzalesGabe:
### Describe the bug.

When the 12 hour clock is enabled, the lock screen clock shows a leading zero for single-digit hours, and "12:XXam" is incorrectly shown as "00:XXam" 

### Reproduction

Make sure 12h clock is enabled
check lockscreen

### Anything else?

the leading zero issue is also present in the status bar clock, like the official firmware. If this is actually a system limitation, I understand it might not be fixable within Momentum Firmware

WillyJL:
Personally I find 00:xx am to be more obvious when wanting to indicate that it is just past midnight. 12:xx am is ambiguous whether it means midnight or midday imo
Is there some standard saying it should be 12:xx am?

GonzalesGabe:
> Personally I find 00:xx am to be more obvious when wanting to indicate that it is just past midnight. 12:xx am is ambiguous whether it means midnight or midday imo Is there some standard saying it should be 12:xx am?

First of all that's a very fair point, I get what you mean, I do see that 00:xx am may feel more explicit/clear for differentiating between midnight and midday, but in a traditional 12 hour clock format,

12:XX AM is midnight
12:XX PM is noon
This is widely accepted in many regions, especially the US, while 00:XX is more associated with 24 hour clocks. It just kind of feels out of place to me. Would it be possible to offer a toggle/customization option so users can choose their preferred midnight formatting (12:XXam vs 00:XXam) so it would cater to both preferences

PR #341, opened 2025-01-01
Clock: 12 hour "midnight format"

956MB:
# What's new

I agree with @GonzalesGabe in #317 that this could probably be an option for users to select their preference on how the 12hr clock is displayed.

Adds the options `12:XX` and `00:XX` in `Settings > System > Midnight Format` to display the preferred format past midnight on all clocks (Desktop, Main Menu MNTM style, Lock screen and Nightstand clock app). "12:30 AM" -> "12:30" OR "00:30".

*"Midnight Format" could also be changed to something like "12AM Format" to eliminate the scrolling.*

#### 12:XX

image (11)

#### 00:XX

image (14)

-----
# For the reviewer


WillyJL (changes requested):
RTC is "sacred" xD

its limited and will persist between firmwares, bad idea to add things here if not also added to OFW

i'd say propose to OFW and see what they think, if they shut it down then we can just use momentum settings

applications/main/momentum_app/scenes/momentum_app_scene_interface_general.c
```diff
@@ -0,0 +1,61 @@
...
+
+void momentum_app_scene_interface_general_var_item_list_callback(void* context, uint32_t index) {
+    MomentumApp* app = context;
+    view_dispatcher_send_custom_event(app->view_dispatcher, index);
+}
+
+static void momentum_app_scene_interface_general_midnight_format_changed(VariableItem* item) {
+    MomentumApp* app = variable_item_get_context(item);
+    bool value = variable_item_get_current_value_index(item);
+    variable_item_set_current_value_text(item, value ? "00:XX" : "XX:00");
```
WillyJL: im assuming this is supposed to be
```suggestion
    variable_item_set_current_value_text(item, value ? "00:XX" : "12:00");
```

WillyJL (approved):

PR #342, opened 2025-01-02
CLI: Add 'clear' command and command suggestion

dexvleads:
# What's new

- Introduced a new CLI command 'clear' (alias 'cls') to clear the terminal screen.
- Enhanced command not found feedback by suggesting similar commands based on user input.
- Added a function to calculate string distance for better command matching.

-----
# For the reviewer


applications/services/cli/cli.c
```diff
@@ -208,6 +209,31 @@ static void cli_execute_command(Cli* cli, CliCommand* command, FuriString* args)
...
+
+static void cli_find_similar_command(Cli* cli, const char* input, FuriString* suggestion) {
+    size_t min_distance = SIZE_MAX;
+    furi_string_reset(suggestion);
+
+    CliCommandTree_it_t it;
+    for(CliCommandTree_it(it, cli->commands); !CliCommandTree_end_p(it); CliCommandTree_next(it)) {
+        const char* cmd_name = furi_string_get_cstr(*CliCommandTree_ref(it)->key_ptr);
+        size_t distance = cli_string_distance(input, cmd_name);
+        if(distance < min_distance && distance <= 3) {
```
WillyJL: so only 3 character difference? wouldnt it be better to base this number off of the length of the input?
the longest commands that would be actually used are `free_blocks` 11 chars and `factory_reset` 13 chars, while shortest are bt, ir, js, top, src, chat... maybe `strlen(input) / 2` would be better? round up too, so 5 chars would allow 3 char difference, and 10 chars would allow 5 char difference

applications/services/cli/cli.c
```diff
@@ -208,6 +209,31 @@ static void cli_execute_command(Cli* cli, CliCommand* command, FuriString* args)
     }
 }
 
+static size_t cli_string_distance(const char* s1, const char* s2) {
+    size_t len1 = strlen(s1), len2 = strlen(s2);
```
WillyJL: this is iterating over the strings twice per call, and called for every command. could get away with iterating once and stopping when finding null bytes

applications/services/cli/cli.c
```diff
@@ -245,9 +271,21 @@ static void cli_handle_enter(Cli* cli) {
     } else {
         furi_check(furi_mutex_release(cli->mutex) == FuriStatusOk);
         cli_nl(cli);
-        printf(
-            "`%s` command not found, use `help` or `?` to list all available commands",
-            furi_string_get_cstr(command));
+        FuriString* suggestion = furi_string_alloc();
+        cli_find_similar_command(cli, furi_string_get_cstr(command), suggestion);
+        
+        if(furi_string_size(suggestion) > 0) {
```
WillyJL: not sure if there is any actual performance benefit, but `furi_string_empty()` seems more fitting here, and might even avoid calculating length (though i think furi string keeps track internally of its length)

Willy-JL (changes requested):

Willy-JL (approved):

PR #345, opened 2025-01-08
added an example js script to show how to implement badusb on an actu…

mqxs21:
…al app (chrome)

# What's new

- Added an example js script on how badusb can be used with chrome.
- This allows new users to more easily create their own scripts by seeing an actual example
-----
# For the reviewer
-I've tested this on multiple Windows PCs
-Also works on older Windows versions as well


WillyJL:
personally i feel like this one is a bit too specific to be included as an example. maybe would be better to share in our discord, in #script-sharing?

PR #349, opened 2025-01-12
Add support for some newer DuckyScript features to BadKb

x10102:
# What's new

Adds support for a couple of new commands to the BadKB scripting language.

| Command                                                                                                                                                    | Description                                                               | Notes                                                                                                                                                                                 |
|------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `CAPS [ON\|OFF]`                                                                                                                       | Ensures that CapsLock is on/off based on the keyboard's current LED state | Similar function to DS `$_CAPSLOCK_ON`                                                                                                                                                |
| `RANDOM_LETTER`<br> `RANDOM_LOWERCASE_LETTER`<br> `RANDOM_UPPERCASE_LETTER`<br> `RANDOM_NUMBER`<br> `RANDOM_SPECIAL`<br> `RANDOM_CHAR` | Generates a random character and sends it to the host                     | Equivalent function to the DS random keystroke injection features                                                                                                                     |
| `VOLUME [0-1]`<br> `BEEP [FREQ] [MS]`                                                                                                  | Plays a tone using the integrated buzzer                                  | This one is a bit silly and doesn't have a DS equivalent, but can be used as an audio prompt before `WAIT_FOR_BUTTON_PRESS` or to notify that a particularly long script has finished |

----
# For the reviewer


WillyJL:
same as #372:
> unfortunately this came in at a bad time. badkb was behind on a few OFW refactors, since those refactors did not add any functionality that badkb didnt already have, but it meant that it was quite differently structured. i just spend the last 3 days rewriting badkb on top of ofw's "new" badusb structure, so now this pr will probably need some reworking.

> also, i would prefer if you tried to contribute this to official firmware first, everything that gets added downstream here becomes an additional maintenance burder when inevitably ofw will change things once again. if they reject these additions, then i will have a look and possibly accept them here

i know this pr didnt really come in at that bad time, but it got to this bad time due to me not looking at it sooner, thats my bad.

still, i would prefer if you tried to contribute this to official firmware first. if declined, i will accept it here

x10102:
Understood. I'll make a PR on the OFW repo when I have the time.

PR #350, opened 2025-01-12
IR: Easy Learn

jaylikesbunda:
# What's new

### Easy Learn™ (if it's not obvious the ™ is a joke)
- Arthritis-Friendly: Pre-named buttons, no typing needed
- Fast: Learn an entire remote like a lot quicker
- Skip: Don't waste time on buttons you'll never use
- No Thinking Required (hopefully): We pick the names, you just press the buttons
- Smart: Learn only the missing buttons from existing remotes

    #### 9 out of 10 thumbs prefer Easy Learn™

### Preview
Screenshot-20250112-224724
image

-----
# For the reviewer


applications/main/infrared/infrared_app.h
```diff
@@ -24,4 +24,5 @@ typedef struct InfraredApp InfraredApp;
 typedef struct {
     FuriHalInfraredTxPin tx_pin;
     bool otg_enabled;
+    bool easy_mode;
```
WillyJL: unfortunately this means it will be incompatible with settings files from other firmwares, and reset to default settings when switching between them. same goes for upgrading from older version of the firmware.

not saying it should be removed, but something to keep in mind. ill think about it and decide tomorrow when i'll review the rest of the PR

jaylikesbunda: could add backwards compat to fallback with old versions?

jaylikesbunda: this is what i've ended up deciding to do, i've bumped the ir settings version to 2 and it will migrate version 1 files on upgrade. bit of extra code but definitely an improvement

WillyJL: All other infrared apps use this settings file to know to use the external modules, and they use the same struct and version from the main app but not the same load logic, so those would lose settings on open if the user opens one of those first and not the main infrared app.
And the easy learn setting doesn't apply to them anyway... it's just stupid that they chose to use saved struct, yeah it's quick to implement but a nightmare for adding new things.
I guess could expose the load and save functions that handle older versions and make all the other infrared apps use it. Can't put it into the API, because it's an external app, easiest solution i see is to put the load and save functions into the .h file, so other apps can use it by just importing the .h file and it gets compiled into them aswell, even if the source code is elsewhere.
Still means that when switching back to another firmware settings will be lost, but atleast on momentum it would be fine. It's the same for desktop settings anyway

WillyJL: working on this now

jaylikesbunda: > working on this now

true 10x dev thank you 🙏
will be able to do any fixes if needed later on today

WillyJL (approved):
love it :100:

jaylikesbunda:
> love it :100: 

love your rewrite 😭❤️

Avaxthorn (commented):
[ReadMe.md](https://github.com/user-attachments/files/19574959/ReadMe.md)

PR #351, opened 2025-01-13
IR: yapper learn icon

jaylikesbunda:
# What's new

cleanup learn icon and add yapper (thanks to @the1anonlypr3)

Screenshot-20250113-234343

-----
# For the reviewer


the1anonlypr3:
<3

WillyJL:
Should be in momentum asset pack, default assets should stay close to stock feel. There was already a yappy ir learn icon but maybe this one is better, replace that one instead and we'll see which one to keep

WillyJL:
Ah I see that before putting yappy in you cleaned up the icon in general. Maybe the cleaned up flipper one could also be PRd to official and we can keep it here in default assets, and yappy one can go in momentum pack

the1anonlypr3:
omds for sure ill PR ofw with the cleaned up icon rn

PR #354, opened 2025-01-15
Main Menu: Adds info toggles to PS4/Vertical/MNTM

956MB:
Depends on some code from #341 to present the clock as `12:34` or `00:34` in the styles.

Adds more info to a few of the menu styles that can fit it, as well as the option to toggle any/all of it on or off. The `Show ...` options in `MNTM > Interface > Main Menu`lock for the other styles that aren't affect (for now), and displays messages for why it's locked.

Also small fix to the MNTM style making the name text scroll if a device name is max length. Like "LOOOOONG".

##### PS4: `name`, `level`, `time`, `battery`

PS4

##### Vertical: `time`, `battery`

Vertical

##### MNTM: `name`, `level`, `time`, `battery`, `5v`

https://github.com/user-attachments/assets/87a2e519-242e-4150-9046-f878c8bfc834

##### Locked settings

Locked

-----
# For the reviewer


WillyJL:
on hold based on #341

PR #356, opened 2025-01-15
New prank badusb

wutchdugsinrl:
# What's new

- [Describe changes here]

-----
# For the reviewer


kitsunehunter:
The porn is good but is there a way to view the video on the flipper screen? Having porn viewable directly on the flipper would be a valid usecase for the majority of flipper users.

kitsunehunter:
@wutchdugsinrl The porn is good but is there a way to view the video on the flipper screen? Having porn viewable directly on the flipper would be a valid usecase for the majority of flipper users.

WillyJL (approved):
xiaohongshu approved :1st_place_medal:

PR #357, opened 2025-01-17
Main Menu: Fixes scrollable text in CoverFlow

956MB:
In #314 

- `elements_scrollable_text_line_centered` was using 0 on scroll, so text wider than the screen would not scroll. Uses `menu_scroll_counter` to fix it.
- Also makes the width 124 to give a 1px margin so the text isn't touching the outside lines

https://github.com/user-attachments/assets/41af2f3e-8c7c-4915-a1cb-ad5f49da5073

-----
# For the reviewer


WillyJL (approved):
nice catch!

PR #363, opened 2025-01-25
GUI: Add marquee 'Text Scroll' option

956MB:
- Adds the tiny `Text Scroll` option under `MNTM > Interface > General` to either use the standard scroll, or a new marquee style back and forth on any scrollable text.
- `elements_scrollable_text_line_centered` renamed to `elements_scrollable_text_line_ex` to be more general. Has `bool marquee`.

https://github.com/user-attachments/assets/05386944-4782-4124-8d44-a2213062101f

https://github.com/user-attachments/assets/79489fc4-ea9e-4f7b-b52b-0b266c84052e

https://github.com/user-attachments/assets/ec1dead6-4b4a-4398-a558-2acc99fa618a

-----
# For the reviewer


WillyJL (changes requested):
i remember discussing the function signature with you but it didnt hit me until seeing the diff in front of me... maybe it would be better to implement this internally to the function? this way the api is the same, and it doesnt require changing all usages of it. and furthermore, it will work on third party apps without recompiling or changing the code at all

the centered function has a different purpose so it had to be separate, but marquee would make sense to apply everywhere i think

WillyJL (approved):
LGTM!

PR #367, opened 2025-02-02
README: Logo light/dark, adaptive stargazers SVG, other cleanup

956MB:
- Adds light and dark versions of the Momentum Firmware logo that adapts to theme
- Adds `?variant=adaptive` to the stargazers url to do the same with the theme
- Misc removal of cluttered line breaks and other `:` / `|`


WillyJL (approved):
dope!

PR #372, opened 2025-02-08
Added support for constants and variables definition to BadKB

fpacenza:
# What's new

- Added the possibility to define constants via the `DEFINE` command
- Added the possibility to define variables via the `VAR` command

### DEFINE
`DEFINE NAME VALUE`
or
`DEFINE #NAME VALUE`

where

`DEFINE` denotes the start of a constant definition
`NAME` or `#NAME` is the label or key used to locate usage within your payload
`VALUE` is the value to replace matching instances of `NAME` throughout your payload

### VAR
`VAR $NAME = VALUE`

where

`VAR` denotes the start of a variable definition
`$NAME` is the variable name used to locate usage within your payload
`VALUE` is the value to replace matching instances of `NAME` throughout your payload

## Payload Example
`DEFINE CONSTANT constant value`
`DEFINE #CONSTANT constant value using # symbol`
`VAR $variable = my variable value`

`STRINGLN this is the value of the constant: CONSTANT`
`STRINGLN this is the value of the #constant: #CONSTANT`
`STRINGLN this is the value of the variable: $variable`

-----
# For the reviewer


WillyJL:
unfortunately this came in at a bad time. badkb was behind on a few OFW refactors, since those refactors did not add any functionality that badkb didnt already have, but it meant that it was quite differently structured. i just spend the last 3 days rewriting badkb on top of ofw's "new" badusb structure, so now this pr will probably need some reworking.

also, i would prefer if you tried to contribute this to official firmware first, everything that gets added downstream here becomes an additional maintenance burder when inevitably ofw will change things once again. if they reject these additions, then i will have a look and possibly accept them here

fpacenza:
Alright! I make a pull request on the OFW, and it’s awaiting review.

WillyJL:
closing as you have opened https://github.com/flipperdevices/flipperzero-firmware/pull/4141 and we could merge that downstream here in case.

PR #379, opened 2025-03-01
Main Menu: Refine CoverFlow menu style

956MB:
#### Before

Original

#### After

Edits

-----
# For the reviewer


WillyJL (approved):
looks nice!

PR #421, opened 2025-05-23
MNTM for Kiisu

WillyJL:
This branch includes small changes and additions specific to the [Kiisu board](https://kiisu.io/). It pulls updates from [Kiisu firmware](https://github.com/kiisu-io/kiisu-firmware) (mostly cosmetic changes such as Kiisu mascot instead of the Dolphin, though this can be reverted with the Flipper asset pack that comes included) and includes a few Kiisu specific apps. We recommend running this build of Momentum on Kiisu boards, even if regular Momentum builds will work too (for the most part).

- Merged changes from [Kiisu firmware](https://github.com/kiisu-io/kiisu-firmware) (fork of OFW with small changes to better support the [Kiisu board](https://kiisu.io/)), adapted for Momentum
- Unlikely ever be official in Momentum releases, for now it stays as a branch here
- The first comment in this thread will always have latest binaries available, and they are also available on the webupdater by selecting the branch option and choosing kiisu-mntm branch (this is why the PR, there is no intention to merge this)


Unc0v3RGD:
Looks really nice, i have been using it in a real kiisu v4b and performs better than the original mntm dev, keep up with the good work!

tototo31:
It appears that the battery readout no longer works correctly. Seems fine on official kiisu. I’ll look into this and provide further info as I have time

WillyJL:
> It appears that the battery readout no longer works correctly. Seems fine on official kiisu. I’ll look into this and provide further info as I have time

It was set to always 100 deliberately as the v4a has a non functional battery management system, in order to get around the annoying warning of dead battery. I'll see about making it only do that if BMS fails, so it can still work on v4b.

WillyJL:
As of 6e810553ea1243a32747539a70a4e6891a8e3807 the base Flipper icons and animations are moved to a "Flipper" asset pack.
By default this branch uses mostly the same graphics as the Kiisu stock firmware, if you don't like them there's the "Flipper" asset pack to get almost same as Momentum on Flipper, and "Momentum" asset pack to get yappy of course.

PR #436, opened 2025-07-16
MNTM Settings: Add Skip Sliding Animations option for Lockscreen

aaronjamt:
# What's new

- The cover animation takes time and blocks input while animating
  - This blocks using desktop keybinds after unlocking, until the animation finishes
- This PR adds an option to skip those animations via the Momentum Settings app
  - Momentum -> Interface -> Lockscreen
  - A new option is added: Skip Sliding Animations

-----
# For the reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   - UL: Nero Radio static parse and display more data (by @xMasterX)
   - UL: Marantec protocol implement CRC verification display and add manually support (by @xMasterX & @li0ard, original code by @Skorpionm)
   - UL: Keeloq Comunello add manually support (by @xMasterX)
+- MNTM Settings: Add Skip Sliding Animations option to skip cover animations (by @aaronjamt)
 
 ### Updated:
 - Apps:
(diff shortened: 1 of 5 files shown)


WillyJL:
Nice! Just curious, why separate options for lock and unlock? Feel like one that does both would be fine, but maybe I just can't think of the right reason

aaronjamt:
> Nice! Just curious, why separate options for lock and unlock? Feel like one that does both would be fine, but maybe I just can't think of the right reason 

The main reason was that I did the unlock piece (which is the one I care more about) first, then realized I might as well add the lock option. I guess you could argue it gives the user more flexibility? I don't see a strong reason either way, I'd be happy to combine them if you want.

WillyJL:
Yeah I think combining would be better. I see this being used also with the transparent option, in which case the slide animation is not shown at all so having the option of skipping it is good, and don't see a reason why you would want fast unlock but slow lock

aaronjamt:
> Yeah I think combining would be better. I see this being used also with the transparent option, in which case the slide animation is not shown at all so having the option of skipping it is good, and don't see a reason why you would want fast unlock but slow lock

Done, and I also renamed it to "Skip Sliding Animation" since that's a bit more descriptive.

Edit: built successfully on my machine, uploading to Flipper to test now.

Edit 2: Can confirm that it works. Somehow I didn't manage to break it this time :)

WillyJL (approved):

PR #438, opened 2025-07-17
Desktop: Fix lock screen hang

aaronjamt:
# What's new

- Randomly, but fairly often, pressing the Up button on the lock screen will show the pin dialog, but will lock up and not accept input for anywhere from 1-10 seconds. 
- This PR fixes this issue by removing the additional delay from the notification sequence.
  - This delay does not exist in OFW.

-----
# For the reviewer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 ### Fixed:
 - Bad KB: Fix modifier keys with HOLD/RELEASE commands (by @WillyJL)
+- Desktop: Fix lock screen hang (#438 by @aaronjamt & @WillyJL)
 
 ### Removed:
 - Nothing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 
 ### Fixed:
 - Bad KB: Fix modifier keys with HOLD/RELEASE commands (by @WillyJL)
-- Desktop: Fix lock screen hang (#438 by @aaronjamt & @WillyJL)
+- Desktop: Fix lock screen hang (#438 by @aaronjamt)
 
 ### Removed:
 - Nothing
--- a/applications/services/desktop/scenes/desktop_scene_locked.c
+++ b/applications/services/desktop/scenes/desktop_scene_locked.c
@@ -102,7 +102,7 @@ bool desktop_scene_locked_on_event(void* context, SceneManagerEvent event) {
         case DesktopLockedEventUpdate:
             if(desktop_view_locked_is_locked_hint_visible(desktop->locked_view)) {
                 notification_message(
-                    desktop->notification, &sequence_display_backlight_off_delay_1000);
+                    desktop->notification, &sequence_display_backlight_off);
             }
             desktop_view_locked_update(desktop->locked_view);
             consumed = true;
--- a/applications/services/desktop/scenes/desktop_scene_locked.c
+++ b/applications/services/desktop/scenes/desktop_scene_locked.c
@@ -101,8 +101,7 @@ bool desktop_scene_locked_on_event(void* context, SceneManagerEvent event) {
             break;
         case DesktopLockedEventUpdate:
             if(desktop_view_locked_is_locked_hint_visible(desktop->locked_view)) {
-                notification_message(
-                    desktop->notification, &sequence_display_backlight_off);
+                notification_message(desktop->notification, &sequence_display_backlight_off);
             }
             desktop_view_locked_update(desktop->locked_view);
             consumed = true;
--- a/applications/services/desktop/scenes/desktop_scene_locked.c
+++ b/applications/services/desktop/scenes/desktop_scene_locked.c
@@ -102,7 +102,7 @@ bool desktop_scene_locked_on_event(void* context, SceneManagerEvent event) {
         case DesktopLockedEventUpdate:
             if(desktop_view_locked_is_locked_hint_visible(desktop->locked_view)) {
                 notification_message(
-                    desktop->notification, &sequence_display_backlight_off);
+                    desktop->notification, &sequence_display_backlight_off_delay_1000);
             }
             desktop_view_locked_update(desktop->locked_view);
             consumed = true;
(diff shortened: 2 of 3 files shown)


aaronjamt:
I've been experiencing this issue for several months and finally decided to troubleshoot it. When the issue happens, the backlight stays off until it unfreezes, which led me to [`desktop_scene_locked.c`](https://github.com/Next-Flip/Momentum-Firmware/blob/b423f90b5ccc6228b60f31fdd3ab19bc0ed380f3/applications/services/desktop/scenes/desktop_scene_locked.c#L104-L105). It calls `notification_message`, passing `&sequence_display_backlight_off_delay_1000` which waits 1 second and then turns off the backlight. By adding a timeout of ~500ms to the [`furi_message_queue_put` in `notification_message`](https://github.com/Next-Flip/Momentum-Firmware/blob/dev/applications/services/notification/notification_app_api.c#L12), then rebooting the Flipper and immediately trying to unlock it (or locking/unlocking until it freezes) with a debugger connected, I was able to find that the NotificationSrv is [waiting for that 1-second delay](https://github.com/Next-Flip/Momentum-Firmware/blob/dev/applications/services/notification/notification_app.c#L330), preventing it from processing any other events and locking up the UI.

aaronjamt:
Here is a video demonstrating the problem:

https://github.com/user-attachments/assets/6412d88d-042e-4822-a746-571058636962

WillyJL:
annoying, the event is being sent asynchronously so notification service should wait and turn off later while desktop continues, but the actual issue is that notification service has only 8 slots in its message queue, so pressing buttons while notification service is waiting to turn off the backlight will fill the event queue and *that* then blocks the input service which is sending events to turn on the backlight due to pressing buttons. for context, raising that to an unreasonable value like 80 hides the bug too. this bit of code is supposed to run after the unlock prompt was shown, to turn off the display early, so maybe its not needed at boot. i'll see if we can make it turn off instantly at boot, and only make this delayed event fire only after the unlock prompt is hidden.

WillyJL (approved):

WillyJL:
looks like i made a mistake 2 years ago in that function :laughing:

aaronjamt:
> annoying, the event is being sent asynchronously so notification service should wait and turn off later while desktop continues, but the actual issue is that notification service has only 8 slots in its message queue, so pressing buttons while notification service is waiting to turn off the backlight will fill the event queue and _that_ then blocks the input service which is sending events to turn on the backlight due to pressing buttons. for context, raising that to an unreasonable value like 80 hides the bug too. this bit of code is supposed to run after the unlock prompt was shown, to turn off the display early, so maybe its not needed at boot. i'll see if we can make it turn off instantly at boot, and only make this delayed event fire only after the unlock prompt is hidden.

Another idea that could help, if it's possible to implement: maybe there's a way it can only add new events to the queue if they differ from the last added event, so pressing X buttons only results in a single "backlight on" event?

aaronjamt:
> looks like i made a mistake 2 years ago in that function 😆

Happens to the best of us :P

WillyJL:
> so pressing X buttons only results in a single "backlight on" event?

i was thinking about that, but id rather not touch something as critical as the input service. unleashed adding the option for vibro feedback in input service  when pressing buttons is already a bit risky imo

also would potentially mean different behavior from ofw and maybe some repercussions on rgb backlight stuff, idk

WillyJL:
this alternative fix worked for me, if you could test on your end too that would be great. i'll be a while before merging, i merge PRs in bulk.

aaronjamt:
> > so pressing X buttons only results in a single "backlight on" event?
> 
> i was thinking about that, but id rather not touch something as critical as the input service. unleashed adding the option for vibro feedback in input service when pressing buttons is already a bit risky imo

I thought that vibro feature was in OFW? I'll double check but I could've sworn I saw it last time I flashed back.

> also would potentially mean different behavior from ofw and maybe some repercussions on rgb backlight stuff, idk

True, although if it's only exactly identical events, I'm not sure how it could have any negative impact, since it'd only happen when the queue is too full (there would be no "last item" to compare against when the queue is empty) and would effectively just be executed as a no-op `x = x;`

aaronjamt:
> this alternative fix worked for me, if you could test on your end too that would be great. i'll be a while before merging, i merge PRs in bulk.

Will do, I'll give it a try when I'm back home.

aaronjamt:
> this alternative fix worked for me, if you could test on your end too that would be great. i'll be a while before merging, i merge PRs in bulk.

That does appear to solve the main issue, but I believe I have also had the lockup happen while the Flipper is already booted after locking then unlocking it. That case hasn't happened in a while, so I could be misremembering, or it could be very rare, but that's why I was focused more on eliminating the delay altogether, rather than just on boot.

I'm not sure what the purpose of the delay here is, as it's not present in OFW and removing it does not have any impact on behavior AFAICT. Even with my original change (switching to `&sequence_display_backlight_off`), the backlight only turns off after the cover is completely closed. Can you explain why there needs to be a delay before turning off the backlight here?

Additionally, I checked the `git blame` on that change you made, and the last time it was modified was in e24cb944ff795387599abfa6d4895ffb1f522957 by xMasterX, but the changelog message was:
```
Desktop: Temp fix for old backlight bug when locking by holding up arrow
```
It appears your alternative fix is effectively just reverting that commit, which I worry might result in a regression bringing back whatever that "old backlight bug" was.

WillyJL:
hm, you're right. git blame was telling me the function was added by me when i jumped to it and quickly looked at that commit's changes, and since our locked view code is quite different in some parts i thought i had made that change and didnt remember.

the delay was added back in xtreme, i think it was something clara noted that it seemed a bit weird for it to turn the screen off immediately as the prompt disappeared. anyway its barely relevant, better to just get rid of it

WillyJL (approved):

aaronjamt:
It seems this didn't entirely solve the problem, I just had a 10-second lockup after leaving it idle for a bit and letting it auto-lock. Will have to troubleshoot and submit a new PR if I figure it out.

At least it seems to have fixed the most consistent/obvious cases.

PR #439, opened 2025-07-17
BT Remote: Add Rename Option, simplify Bad KB BLE profile

### LINKED ISSUE #410: Change BLE Remote Name
### Description of the feature you're suggesting.

It would be nicer to have the ability to change the ble remote name, from Control + Flipper name to something custom. I imagine the app having above Bluetooth Unparing, a menu called, Remote Name, which when pressed it would allow you to input a whole new name, without control without anything, so something completely custom. 
Thank you 

### Anything else?

_No response_

aaronjamt:
# What's new

- This applies to the Bluetooth Remote app.
- Adds an option to rename the advertised Bluetooth device.

Addresses #410.

-----
# For the reviewer


lib/ble_profile/extra_profiles/hid_profile.c
aaronjamt, line 1: This was the only way I could find to be able to replace the Flipper's name, rather than just the prefix. I can also revert this and only make it change the "Control" prefix, although that feels less useful to me.

WillyJL: This is part of firmware and available for apps via api (kinda, needs fap_libs, but still applies), so cannot change its api to keep ofw compatibility. Copy paste the custom profile used in badkb, it is the same ble profile but with parameter for full name.

WillyJL (changes requested):
As mentioned in review comment

lib/ble_profile/extra_profiles/hid_profile.c
aaronjamt: > This is part of firmware and available for apps via api (kinda, needs fap_libs, but still applies), so cannot change its api to keep ofw compatibility.

Makes sense, good to know. I'll try to be more careful about that in the future.

> Copy paste the custom profile used in badkb, it is the same ble profile but with parameter for full name.

I believe I have done that, hopefully [this](https://github.com/Next-Flip/Momentum-Firmware/pull/439/files#diff-9ea367af10dc11f784c94b253b0b509eecbd56f56f3ba6c8eb0a1ad7545f89f9R6-R49) is what you meant?

aaronjamt:
Should I combine these commits into one (or at least remove the separate "revert changes to firmware" commit) and force-push or are "revert" commits okay to be merged into the main repo?

WillyJL:
I typically squash merge pull requests, so it doesn't matter anyway :) thanks, will review soon

aaronjamt:
> I typically squash merge pull requests, so it doesn't matter anyway :) thanks, will review soon

Sounds good, thanks!

lib/ble_profile/extra_profiles/hid_profile.c
WillyJL: huh, you outsmarted me... (not that it takes much lol)

never occurred to me that you can just replace the get_cap_config method, i just copy pasted the whole service file into badkb back then. hope you dont mind me using your solution in badkb too xD

nicely done!

WillyJL: think this can be made even simpler by having a custom profile params struct with the basic params struct at the beginning, and calling the firmware's profile with that struct, then editing the name afterwards. also makes sure it wont bitrot over time if the firmware profile ever changes and we forget to change the equivalent here. gonna try that shortly

aaronjamt: > hope you dont mind me using your solution in badkb too xD

Not at all! This is why I love open-source: one man's spaghetti-code is another man's solution... or something like that, anyways.

> think this can be made even simpler by having a custom profile params struct with the basic params struct at the beginning, and calling the firmware's profile with that struct, then editing the name afterwards. also makes sure it wont bitrot over time if the firmware profile ever changes and we forget to change the equivalent here. gonna try that shortly

Just reading this on my phone, I'm not entirely sure what you're proposing here, but if you have a better idea feel free to edit this PR. I believe the profile params is just a void* so I used it to point to the char* for the device name (avoiding needing a params struct altogether), so I didn't think that's risking any future firmware change issues, but you'd know better about that than me. I'll trust your judgement in any case!

WillyJL:
also no need to worry about rebasing, it ends up squashed anyway so i merge into pr branches just to test on latest changes and to update the changelog file, and that wont cause issues for squashing. if i have some local changes im yet to push to the pr branch tho, rebasing gets in the way :)

aaronjamt:
> also no need to worry about rebasing, it ends up squashed anyway so i merge into pr branches just to test on latest changes and to update the changelog file, and that wont cause issues for squashing. if i have some local changes im yet to push to the pr branch tho, rebasing gets in the way :)

Sounds good, won't rebase again. I just like seeing a nice tidy graph in VSCode, lol

WillyJL:
i think ill edit this a bit more than i wanted to, wanna add a setting file so it remembers, so this profile will need to be moved outside the rename scene and be used from the beginning, similar to badkb.

aaronjamt:
> i think ill edit this a bit more than i wanted to, wanna add a setting file so it remembers, so this profile will need to be moved outside the rename scene and be used from the beginning, similar to badkb.

Would you rather I do that so you don't have to? I didn't mean to add a ton of work to your plate.

WillyJL:
not at all, i want some fun too :D working on it rn

applications/system/hid_app/scenes/hid_scene_rename.c
```diff
@@ -0,0 +1,82 @@
...
+    Hid* app = context;
+
+    view_dispatcher_send_custom_event(app->view_dispatcher, HidSceneRenameEventTextInput);
+}
+
+void hid_scene_rename_popup_callback(void* context) {
+    Hid* app = context;
+
+    view_dispatcher_send_custom_event(app->view_dispatcher, HidSceneRenameEventPopup);
+}
```
WillyJL, line 20: a little note on this, it worked totally fine how you had it, but doing things inside of GUI callbacks can come back to hit you in the face; especially scenemanager stuff, changing scenes means switching GUI view, and doing that inside of a GUI callback could mean a deadlock. wasn't the case here, but it can become a problem later on, i had this in ble spam, started having issues a year after writing the code for it lol. generally speaking try to do control flow and larger tasks on the app thread, and with view dispatcher the events are handled on the app thread sequentially.

WillyJL (approved):

applications/system/hid_app/scenes/hid_scene_rename.c
aaronjamt: Noted, thanks for the advice!

lib/ble_profile/extra_profiles/hid_profile.c
WillyJL: i meant something like 
```c
typedef struct {
    BleProfileHidParams base;
    char name[FURI_HAL_BT_ADV_NAME_LENGTH];
} BleProfileHidExtParams;
```
so the wrapper profile's functions could call the base profile with a base params struct. but turns out bt remote wasnt using any params anyway, so i did away with that, and just made it call the original profile's functions with NULL params

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   - UL: Nero Radio static parse and display more data (by @xMasterX)
   - UL: Marantec protocol implement CRC verification display and add manually support (by @xMasterX & @li0ard, original code by @Skorpionm)
   - UL: Keeloq Comunello add manually support (by @xMasterX)
+- BT Remote: Add Rename Option, simplify Bad KB BLE profile (by @aaronjamt & @WillyJL)
+- MNTM Settings: Add Skip Sliding Animations option for Lockscreen (by @aaronjamt)
```
aaronjamt, line 10: I think you accidentally added this line to the wrong branch

WillyJL: if i dont add it here it gets a conflict when merging later, since theyre close together. usually i would merge one first, then merge dev into the other, then merge the other pr; but i think adding the other line to the second pr, and merging one after the other should avoid the conflict, so im trying it for the first time lol

aaronjamt: Ah, alright. Sounds complicated but if it works, it works lol

WillyJL: welp that didnt work, nevermind lol

WillyJL: thank you!

aaronjamt: Figured I'd help since soon you're going to take all my credit for this PR :laughing:

PR #440, opened 2025-07-18
Subghz fix

Nhscan:
# What's new

- [Describe changes here]

-----
# For the reviewer


WillyJL:
This is not a fix, frequency analyzer does not work correctly on external boards because they lack the extra hardware components for it to work reliably and accurately. It was removed intentionally.

PR #451, opened 2025-07-29
Replace commas with dashes to avoid naming errors

Stichoza:
According to lab.flipper.net, files are not allowed to contain commas. It displays an error right when you click rename on files. To avoid this error, it would be good to replace comma with dashes.

## Example

image

image

# What's new

- Replaced commas with dashes

-----
# For the reviewer

### DIFF
--- a/lib/toolbox/name_generator.c
+++ b/lib/toolbox/name_generator.c
@@ -115,7 +115,7 @@ void name_generator_make_detailed_datetime(
             snprintf(
                 name,
                 max_name_size,
-                "%.4d-%.2d-%.2d_%.2d,%.2d,%.2d_%s",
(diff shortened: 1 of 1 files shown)


WillyJL:
What a goofy limitation, can't think of any reason why commas would be an issue, maybe they just forgot to add it

Stichoza:
Yeah, but dashes are more common in datetime strings than commas anyways so idk

WillyJL:
They are for dates, I would've used colons but that's an illegal character. Didn't use dots as they are for the extension. Commas are thin so the text doesn't go off screen too much and also it differentiates from the date and extension

xMasterX:
> What a goofy limitation, can't think of any reason why commas would be an issue, maybe they just forgot to add it

I think mobile app had same limitation at ios, not sure if its still in place

PR #468, opened 2025-09-14
Allow building the firmware with the SubGHz protocol items disabled

user890104:
# What's new

In order to fit large protocol decoders/encoders, some of the others need to be disabled in order to meet the flash size limit. But some of the code relied on them always enabled.

This patch fixes it by including the relevant headers instead of assuming that all SubGHz protocols are enabled at all time.

Tested building with all protocols disabled, SubGHz Read and Read raw crash because of furi_check(). By enabling just raw and bin_raw, both options work again.

No functional change is expected by this commit.

-----
# For the reviewer

### DIFF
--- a/applications/main/subghz/helpers/subghz_gen_info.c
+++ b/applications/main/subghz/helpers/subghz_gen_info.c
@@ -1,6 +1,19 @@
 #include "subghz_gen_info.h"
 #include "../helpers/subghz_txrx_create_protocol_key.h"
-#include <lib/subghz/protocols/protocol_items.h>
+#include <lib/subghz/protocols/princeton.h>
+#include <lib/subghz/protocols/nice_flo.h>
+#include <lib/subghz/protocols/came.h>
+#include <lib/subghz/protocols/roger.h>
+#include <lib/subghz/protocols/linear.h>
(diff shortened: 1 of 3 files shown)


WillyJL:
by "SubGHz protocol items disabled" do you mean you were also commenting the `#include`s for them in `protocol_items.h`? because if so, thats not necessary, you can disable protocols by commenting them from `protocol_items.c`, specifically just removing them from the `subghz_protocol_registry_items` array: this will mean they are not part of the list of protocols used for the firmware, and thus they are not referenced elsewhere in the binary and so they are stripped when linking the firmware, meaning you can fit whatever you needed without deleting files or includes, just commenting them from the subghz_protocol_registry_items array, and making this PR redundant afaict

user890104:
That's right, I just went one step ahead and checked if there are any assumptions that specific protocol headers are always loaded, and found these occurrences. IMO it's cleaner to include only the bits you specifically need, so feel free to close it instead.

WillyJL:
yeah i appreciate the PR but to build with protocols disabled you should just comment them out of the `subghz_protocol_registry_items` array, then the linker can figure out what can be removed from the final binary, and headers stay where they are

PR #494, opened 2025-12-02
Hate Claude, Hate ChatGPT, Hate Deepseek

OrionW06:
# What's new

- New Claude Rules to prevent the use of GenAI in this codebase

-----
# For the reviewer


956MB:
Even people who use LLM's know how to delete files. It doesn't make any sense to have this.

PR #500, opened 2025-12-30
NFC: Handle PPS request in ISO14443-4 layer

WillyJL:
# What's new

- PPS handled in ISO14443-4 layer, should help with SEOS app compatibility

-----
# For the reviewer


aaronjamt:
Just tested with a genuine HID reader, this does fix the firmware-side issues with the Seos app.

PR #501, opened 2026-01-01
NFC: ISO 15693-3 emulation READ_MULTI and GET_BLOCK_SECURITY fixes

### LINKED ISSUE #417: NFC emulate ISO15693-3 Flipper crash and reboot
### Describe the bug.

During NFC ISO15693-3 emulation of a certain access card, FLIPPER crashed and was rebooted. I think I tried all the possibilities: disabling animations, debug mode and everything that can be changed in the settings. The strange thing is that the emulation of the ISO15693-3 card works fine, without hanging the Flipper only on XTREME software. 
In order for flipper to crash, you need to to simulate a terminal reading this card, you can use an android phone and Taginfo program. When trying to read with the phone, the same situation occurs as with the terminal - Flipper crashes and restarts. Greetings. Best Regards

### Reproduction

1. upload the NFC file from the flipper attachment: sd card/nfc;
2. turn on the flipper;
3. select NFC;
4. select the previously uploaded “tech” file;
5. emulate IT

[tech.zip](https://github.com/user-attachments/files/20344315/tech.zip)

aaronjamt:
# What's new

Fixes some errors in ISO 15693-3 emulation, namely READ MULTI command returning 1 block more than expected, and crashing with some unexpectedly large command responses (issue #417).

- Fixes the bug in issue #417 (hopefully)
- Fixes READ MULTI returning 1 more block than expected (ISO 15693-3 READ MULTI has the block count in the command as 1 less than what should be returned, eg command with block count 0 should return 1 block, this was previously returning 2 blocks instead)
- Allocates enough response buffer size for a full 256 block response to BLOCK SECURITY command
- Fixes READ MULTI so that asking for more blocks than can fit into the response buffer will return an error instead of crashing

-----
# For the reviewer


WillyJL:
this makes sense to me, but im drafting until i can check on a real card in a few days what it does with a READ MULTI command larger than it can handle, if it just errors or if it returns less blocks than asked.

aaronjamt:
Given the positive response in #417, I'd say this is probably ready for undraft and merge.

WillyJL:
> this makes sense to me, but im drafting until i can check on a real card in a few days what it does with a READ MULTI command larger than it can handle, if it just errors or if it returns less blocks than asked.

checked with SLIX2, there is no reply if asked for too many blocks in one "read multi" command. changed behavior to mirror that

WillyJL (approved):

PR #514, opened 2026-01-30
UI tweaks in Momentum app

kaya-arro:
# What's new

- Removed '>' symbol from menu items in Momentum app that do not respond to "right" button presses
- Improved spacing and capitalization in Momentum app interface scene
- Small grammar corrections in `ReadMe.md`
- Ran formatter

-----
# For the reviewer


956MB:
I might argue that removing the `>` would make it more confusing from an interaction stand point, and definitely more bare visually. The `OK` to click into something is pretty universal and clearly implied in all the screens as the `>` represents like a "click to go into this page", not really a "`RIGHT` click necessary". I think it's a non issue.

The grammar corrections are in `CODING_STYLE.md`, not any `README.md`, which is an infrequently used non forward facing piece of documentation... so aesthetic changes here aren't really crucial.

Running scripts/lint.py is not necessary, only `./fbt format`. And btw you do not need to delete and create and new PR if you make a change, new commits will be amended onto it.

WillyJL:
what 956 said.

and the formatting changes are not supposed to happen. as he said just `./fbt format` is sufficient and targets files that should be formatted. some of the files you modified here are already formatted by fbt normally, which leads me to believe you have some different configuration somewhere leading to your fbt formatting files differently. all in all nothing in this pr i have interest in merging.

PR #524, opened 2026-02-25
NFC: Add Mifare Ultralight C Write Support

haw8411:
# What's new

-  Added Ultralight C Write functionality. With this, you can now, as long as keys are saved in the dictionary, write to Ultralight C cards. You have 2 options: `Write (Keep Key)` and `Write (Copy Key)`. These 2 options allow you to pick if you want to overwrite the key on the target card; or keep it. This also checks the dictionaries for keys & attempts to authenticate, allowing you to not just write to the card you just read, but other cards given you have its key stored in the dictionaries. For faster writing, it also stores those keys associated with the UID in `/ext/nfc/assets/mf_ultralight_c_key_cache.nfc` so it doesn't have to check and run through the whole dictionary before writing. 

-----
# For the reviewer


haw8411:
Please let me know if you find any bugs in the code, or have any suggestions to improve on it. I've verified multiple times that everything works, including re-flashing my Flipper to OFW, and then compiling and flashing back & testing again. Thanks :)

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -210,12 +230,31 @@ static void nfc_scene_read_and_saved_menu_on_enter_mf_ultralight(NfcApp* instanc
     const MfUltralightData* data =
         nfc_device_get_data(instance->nfc_device, NfcProtocolMfUltralight);
     bool is_locked = !mf_ultralight_is_all_data_read(data);
+    bool is_mfulc = (data->type == MfUltralightTypeMfulC);
```
aaronjamt: Nit: There's probably no reason to make this a separate variable since it's only used in one place

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -291,6 +330,16 @@ static bool nfc_scene_read_and_saved_menu_on_event_mf_ultralight(
                 scene_manager_next_scene(instance->scene_manager, NfcSceneMfUltralightCDictAttack);
             }
             consumed = true;
+        } else if(event.event == SubmenuIndexWriteKeepKey) {
+            s_ulc_write_copy_key = false;
+            FURI_LOG_D("MfULC", "Write menu: KEEP KEY selected");
```
aaronjamt: Nit: Not sure we need extra debug logs for just transitioning to the next scene

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -307,12 +356,342 @@ static NfcCommand
...
+                                    instance->mf_ultralight_c_dict_context.dict_keys_current = 99;
+                                    return NfcCommandContinue;
+                                }
+                            }
+                        }
+                        furi_string_free(uid_tag);
+                        furi_string_free(content);
+                    } else {
+                        storage_file_close(file);
+                    }
```
aaronjamt: Is there a reason we're not using the methods from `flipper_format.h` here? Seems like it'd be easier and cleaner to use that to read/write the cache file.

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -386,7 +792,7 @@ const NfcProtocolSupportBase nfc_protocol_support_mf_ultralight = {
     .scene_write =
         {
             .on_enter = nfc_scene_write_on_enter_mf_ultralight,
-            .on_event = nfc_protocol_support_common_on_event_empty,
+            .on_event = nfc_scene_write_on_event_mf_ultralight,
```
aaronjamt: There's no reason for this, you're just reimplementing `nfc_protocol_support_common_on_event_empty`.

lib/nfc/protocols/mf_ultralight/mf_ultralight.c
```diff
@@ -575,14 +575,18 @@ uint8_t mf_ultralight_get_write_end_page(MfUltralightType type) {
...
+        type == MfUltralightTypeNTAG216 || type == MfUltralightTypeOrigin ||
+        type == MfUltralightTypeMfulC);
 
     uint8_t end_page = mf_ultralight_get_config_page_num(type);
     if(type == MfUltralightTypeNTAG213 || type == MfUltralightTypeNTAG215 ||
        type == MfUltralightTypeNTAG216) {
         end_page -= 1;
     } else if(type == MfUltralightTypeOrigin) {
         end_page = mf_ultralight_features[type].total_pages;
+    } else if(type == MfUltralightTypeMfulC) {
```
aaronjamt: Nit: Could merge this with the previous condition like:
```
} else if(type == MfUltralightTypeOrigin || type == MfUltralightTypeMfulC) {
    end_page = mf_ultralight_features[type].total_pages;
}
```

lib/nfc/protocols/mf_ultralight/mf_ultralight_poller.c
```diff
@@ -738,6 +741,93 @@ static NfcCommand mf_ultralight_poller_handler_request_write_data(MfUltralightPo
...
+        uint8_t target_uid_len = 0;
+        uint8_t target_uid[10] = {0};
+        if(instance->data->iso14443_3a_data) {
+            target_uid_len = instance->data->iso14443_3a_data->uid_len;
+            if(target_uid_len > 10) target_uid_len = 10;
+            memcpy(target_uid, instance->data->iso14443_3a_data->uid, target_uid_len);
+        }
+
+        bool auth_ok = false;
+        bool card_dead = false;
```
aaronjamt: I don't think this variable is necessary, you're already breaking from the loop when setting it to true anyways.

aaronjamt (commented):
Just tested with a few of my cards and it works! Just have a few minor things to check.

aaronjamt:
While I understand why you're caching the keys, I'm not sure how useful that would be realistically. If the card's keys change externally then you're going to have to search the dictionary anyways, plus every card would have to search the dict the first time anyways. I don't necessarily have a problem with it, just not sure if it justifies the added code complexity.

haw8411:
Yeah, I was wondering if it might be smart to remove the cache. 

I'll fix what you found :)

haw8411:
@aaronjamt I just fixed what you suggested, and did end up removing the cache.

lib/nfc/protocols/mf_ultralight/mf_ultralight_poller.c
```diff
@@ -738,6 +741,85 @@ static NfcCommand mf_ultralight_poller_handler_request_write_data(MfUltralightPo
...
+    // WriteKeyRequest so the callback can cache the found key and return the skip decision.
+    if(check_passed &&
+       mf_ultralight_support_feature(features, MfUltralightFeatureSupportAuthenticate)) {
+        uint8_t target_uid_len = 0;
+        uint8_t target_uid[10] = {0};
+        if(instance->data->iso14443_3a_data) {
+            target_uid_len = instance->data->iso14443_3a_data->uid_len;
+            if(target_uid_len > 10) target_uid_len = 10;
+            memcpy(target_uid, instance->data->iso14443_3a_data->uid, target_uid_len);
+        }
```
aaronjamt: Now that there's no cache, I don't see any use for this "target_uid" stuff.

aaronjamt (commented):
Looks good, only one thing I noticed.

lib/nfc/protocols/mf_ultralight/mf_ultralight_poller.c
aaronjamt: Also: https://github.com/Next-Flip/Momentum-Firmware/pull/524/changes/BASE..1b8d3f5fe8093adaaa2170597f9a2a871542643d#diff-7bdc6b55a664ebca6f3f293fb324b5538324f252f3e6ebf05680ef8cf823401dR60-R61

haw8411:
Just removed the target_uid stuff

aaronjamt:
Looks good to me! I'll do some testing and verify there's no issues, if not I think this will be ready to merge.

haw8411:
Sounds good, thanks for taking your time to look over this!

aaronjamt (approved):

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -14,14 +14,34 @@ enum {
...
 
+// ULC write key choice - stored as a static so scene_manager_next_scene can't wipe it.
+// Only one write runs at a time so a single static is safe.
+static bool s_ulc_write_copy_key = false;
+
+// Tracks whether mf_ultralight_c_dict_context.dict was allocated by OUR write callback.
+//
+// The stock NfcSceneMfUltralightCDictAttack on_exit calls keys_dict_free() on the dict
+// but leaves the dict pointer non-NULL (dangling). If write on_enter blindly calls
+// keys_dict_free() on that dangling pointer it double-frees and crashes.
```
WillyJL: why not fix the dangling pointer? then we dont need this workaround variable, unless im missing something

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -14,14 +14,34 @@ enum {
...
 };
 
 enum {
     NfcSceneMoreInfoStateASCII,
     NfcSceneMoreInfoStateRawData,
 };
 
+// ULC write key choice - stored as a static so scene_manager_next_scene can't wipe it.
+// Only one write runs at a time so a single static is safe.
+static bool s_ulc_write_copy_key = false;
```
WillyJL: i'd rather this be moved to NfcApp struct, eg after mf_ultralight_c_dict_context as mf_ultralight_c_write_copy_key

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -307,12 +353,148 @@ static NfcCommand
...
+            if(instance->mf_ultralight_c_dict_context.dict) {
+                keys_dict_free(instance->mf_ultralight_c_dict_context.dict);
+                instance->mf_ultralight_c_dict_context.dict = NULL;
+            }
+            // Sentinel: prevent the if(!dict) block from re-opening dicts on the next call.
+            // Without this, after exhaustion dict==NULL resets the state machine into
+            // an infinite loop of re-opening and re-reading both dictionaries.
+            instance->mf_ultralight_c_dict_context.dict_keys_current = 100;
+            s_ulc_dict_owned = false;
+        }
```
WillyJL, line 426: dict_keys_current is used very improperly here. its supposed to store the current key, eg for showing progress in ui. here it is used for deciding which dictionary to open? and what is this unlabelled 1, 2, 100? comments do not excuse wrong variable usage.
use an appropriate variable, an enum would be ideal, and dont store it in an unrelated variable.

i would suggest merging this with the other comment i put: maybe make a NfcMfUltralightCWriteContext struct in nfc_app_i.h that contains both whether to copy or replace the key and the state of which dict we are using, and store an NfcMfUltralightCWriteContext in NfcApp struct.

WillyJL (changes requested):
no other complaints, rest looks good, nice work. will happily merge once these code quality concerns are fixed. thanks for the PR!

(fixing these also increases the likelihood that OFW would accept this as a backported PR, though in this state it relies on PR 4242 (it reworks how the write scene works in nfc app) which is merged here but not in OFW)

aaronjamt:
> (fixing these also increases the likelihood that OFW would accept this as a backported PR, though in this state it relies on PR 4242 (it reworks how the write scene works in nfc app) which is merged here but not in OFW)

How difficult do you think it would be to back-back-port this to pre-4242? If they start looking at PRs again soon that might be an option to try to get this in a bit faster.

haw8411:
Sounds good, I'll go ahead and fix what you found! Thanks :)

WillyJL:
> > (fixing these also increases the likelihood that OFW would accept this as a backported PR, though in this state it relies on PR 4242 (it reworks how the write scene works in nfc app) which is merged here but not in OFW)
> 
> How difficult do you think it would be to back-back-port this to pre-4242? If they start looking at PRs again soon that might be an option to try to get this in a bit faster.

shouldnt be too difficult. iirc, i basically just moved the poller callback from the dedicated scene into the protocol_support mf_ultralight.c and the different ui text into `furi_string_set(instance->text_box_store, ...)` calls.

haw8411:
Just fixed what @WillyJL suggested, let me know if you guys find anything else!

applications/main/nfc/helpers/protocol_support/mf_ultralight/mf_ultralight.c
```diff
@@ -353,14 +335,11 @@ static NfcCommand
     if(mf_ultralight_event->type == MfUltralightPollerEventTypeRequestMode) {
         mf_ultralight_event->data->poller_mode = MfUltralightPollerModeWrite;
         furi_string_reset(instance->text_box_store);
-        // Free any dict handle left open by the read phase before resetting context.
-        // Without this, every write-scene entry leaks a file handle (one per card read).
         if(instance->mf_ultralight_c_dict_context.dict) {
             keys_dict_free(instance->mf_ultralight_c_dict_context.dict);
         }
         instance->mf_ultralight_c_dict_context.dict = NULL;
```
aaronjamt: Looks like you just removed the comment here but you're still checking for the handle and freeing it. I don't think there's any specific reason to remove this comment if you're still going to do this check-and-free; I'd either keep both or remove both.

applications/main/nfc/nfc_app.c
```diff
@@ -42,6 +42,17 @@ static void nfc_app_rpc_command_callback(const RpcAppSystemEvent* event, void* c
 NfcApp* nfc_app_alloc(void) {
     NfcApp* instance = malloc(sizeof(NfcApp));
 
+    // malloc does not zero-initialize. Explicitly zero the ULC context structs so
+    // that any code reading them before the first write (e.g. prepare_view on first
+    // tap) sees clean NULL/false/Idle values rather than garbage heap data.
```
aaronjamt: Nit: nothing else in here is commented, and I think this goes without saying. Maybe just label the section with a short one-liner instead, or remove the comment altogether?

applications/main/nfc/nfc_app.c
```diff
@@ -42,6 +42,17 @@ static void nfc_app_rpc_command_callback(const RpcAppSystemEvent* event, void* c
...
+    instance->mf_ultralight_c_dict_context.dict = NULL;
+    instance->mf_ultralight_c_dict_context.auth_success = false;
+    instance->mf_ultralight_c_dict_context.is_card_present = false;
+    instance->mf_ultralight_c_dict_context.dict_keys_total = 0;
+    instance->mf_ultralight_c_dict_context.dict_keys_current = 0;
+    instance->mf_ultralight_c_write_context.copy_key = false;
+    instance->mf_ultralight_c_write_context.dict_state = NfcMfUltralightCWriteDictIdle;
+
```
aaronjamt: Nit: This would make more sense at the bottom of the function, next to the `instance->iso14443_3a_edit_data`, `instance->file_path`, and `instance->file_name` initialization.

applications/main/nfc/scenes/nfc_scene_mf_ultralight_c_dict_attack.c
```diff
@@ -77,6 +77,15 @@ void nfc_scene_mf_ultralight_c_dict_attack_prepare_view(NfcApp* instance) {
...
     dict_attack_set_type(instance->dict_attack, DictAttackTypeMfUltralightC);
 
+    // Guard: if a previous write phase left a dict handle open, close it now.
+    // Without this, navigating write->back->read->dict-attack would open the same
+    // file twice, corrupting VFS state and causing a ViewPort lockup.
+    if(instance->mf_ultralight_c_dict_context.dict) {
+        keys_dict_free(instance->mf_ultralight_c_dict_context.dict);
+        instance->mf_ultralight_c_dict_context.dict = NULL;
+    }
+    instance->mf_ultralight_c_write_context.dict_state = NfcMfUltralightCWriteDictIdle;
```
aaronjamt: Nit: I'd move the `mf_ultralight_c_write_context.dict_state` update next to each of the places where the `mf_ultralight_c_dict_context.dict` is actually set to NULL.

applications/main/nfc/scenes/nfc_scene_mf_ultralight_c_dict_attack.c
```diff
@@ -77,6 +77,15 @@ void nfc_scene_mf_ultralight_c_dict_attack_prepare_view(NfcApp* instance) {
     // Set attack type to Ultralight C
     dict_attack_set_type(instance->dict_attack, DictAttackTypeMfUltralightC);
 
+    // Guard: if a previous write phase left a dict handle open, close it now.
+    // Without this, navigating write->back->read->dict-attack would open the same
+    // file twice, corrupting VFS state and causing a ViewPort lockup.
+    if(instance->mf_ultralight_c_dict_context.dict) {
+        keys_dict_free(instance->mf_ultralight_c_dict_context.dict);
+        instance->mf_ultralight_c_dict_context.dict = NULL;
+    }
```
aaronjamt, line 87: Is this really necessary? You're already freeing the dict in the `_on_exit` method, so navigating away from this scene should already clean this up.

aaronjamt: An easy way to check would be to add a FURI_LOG in here and then do that sequence of navigation steps and see if you ever that ever shows in the log

applications/main/nfc/nfc_app.c
WillyJL: on flipper, malloc does zero initialize

haw8411: Just fixed it :)

applications/main/nfc/scenes/nfc_scene_mf_ultralight_c_dict_attack.c
haw8411: Yeah, I did end up adding a log in fact and it did end up showing in the logs

aaronjamt: In that case there's no need to keep that log in the final code

WillyJL (approved):
nice work!

haw8411:
Thanks! Excited to see this in the FW soon hopefully :)

Issue #449, opened 2025-07-27
Show more details about signals


Stichoza:
### Describe the enhancement you're suggesting.

I wonder if it's possible to display more detailed information about captured signals. For example, if you run `rtl_433` it will decode more detailed information. For example, it detects models (General-Remote, Kerui-Security, Chuango-Security, etc.), Tri-State data, commands (like Low Battery, Motion Detected, etc.).

On the Flipper Zero, same signals are interpreted simply as "Princeton 24bit", no more info except the hex key and button.

### Anything else?

Example:

```
time      : 2025-07-27 18:25:07
model     : Chuango-Security                       ID        : 218380
CMD       : Low Battery  CMD_ID    : 13
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2025-07-27 18:28:27
model     : Generic-Remote                         House Code: 38906
Command   : 166          Tri-State : XZZ111XXXXZX
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2025-07-27 18:28:48
model     : Generic-Remote                         House Code: 61042
Command   : 54           Tri-State : 1X1XZ10X01ZX
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
time      : 2025-07-27 18:29:24
model     : Secplus-v1   id        : -32154729
ID_0      : 0            ID_1      : 1             Switch-ID : 0             Remote-ID : -32154729     Button-ID : middle        Fixed_Code: 3426789600    Rolling_Code: 1137999014
```

PR #529, opened 2026-03-08
ISSUE-449: Initial pass at creating 4 and 8 bit branches for the decoder feed

herbenderbler:
# What's new

- [ISSUE-449](https://github.com/Next-Flip/Momentum-Firmware/issues/449): I added more contextual information in SubGHz decode and history for Princeton 24-bit by adding Chuango-style command labels along with a dedicated one-line history string.

-----
# For the reviewer

-----
# Additional info

This PR is contingent upon #530 (and/or additional flash reclamation work) landing

### DIFF
--- a/lib/subghz/protocols/princeton.c
+++ b/lib/subghz/protocols/princeton.c
@@ -53,6 +53,8 @@ typedef enum {
     PrincetonDecoderStepCheckDuration,
 } PrincetonDecoderStep;
 
+void subghz_protocol_decoder_princeton_get_string_brief(void* context, FuriString* output);
+
 const SubGhzProtocolDecoder subghz_protocol_princeton_decoder = {
     .alloc = subghz_protocol_decoder_princeton_alloc,
     .free = subghz_protocol_decoder_princeton_free,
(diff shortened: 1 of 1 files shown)


WillyJL:
i like this concept, this is another reason why `get_string_brief` was added here downstream. there is an argument to be had on whether this flash space is better used for this kind of extra info or for more space for more protocols. in any case i cant in good faith consider merging this until #530 is in (or we free some flash some other way, we are critically low).

herbenderbler:
> i like this concept, this is another reason why `get_string_brief` was added here downstream. there is an argument to be had on whether this flash space is better used for this kind of extra info or for more space for more protocols. in any case i cant in good faith consider merging this until #530 is in (or we free some flash some other way, we are critically low).

That makes sense.  While I'll obviously continue tracking PR #530, I'm curious if I could possibly assist by:

- shrinking my own change (ie. shorter strings, fewer labels, optional compile-time toggle, etc.)
- contributing to PR #530 and/or other flash-reduction work

As it seems my work here was received positively, how would you prefer I move forward with it?  Should I keep it open until flash is in a better state, until #530 lands, etc.?

All of that being said, I'd really LOVE to contribute somewhere in Momentum with more impact.  Any guidance you can offer around future contributions and/or this PR would be greatly appreciated ❤️

WillyJL:
from a quick look your code seems fine, i wouldnt worry about shrinking it. i was just mentioning that in general we're low on flash, thats something that i inevitably have to get fixed at some point, i didnt mean that this pr had to be done different.

id say keep this open and if you intended to implement something like this for some other protocol go ahead and add it in here too.

PR #531, opened 2026-03-09
Update GHA patch levels

herbenderbler:
# What's new

I updated the patch level versions of the following GitHub Actions:

```
actions/checkout@v4 => v4.3.1
actions/upload-artifact@v4 => v4.3.1
Codex-/await-remote-run@v1 => v1.13.0
```

My intention for these version bumps is to keep linchpin workflows current for maintenance, dependency updates, and safer, more predictable behavior, without the risk of introducing breaking changes.

-----
# For the reviewer

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
         target: [f7]
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@v4
+        uses: actions/checkout@v4.3.1
         with:
           fetch-depth: 1
           ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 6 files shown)


WillyJL:
thanks for the pr, however the main point for this change, not introducing breaking changes inadvertently, is already solved. the actions versions use semantic versioning, where the major version number gets bumped when breaking changes are introduced. thats why all actions suggest tracking `v4` or something like that, so they can make v4.0.1 and v4.1.2 and so on which can fix bugs and add new features with no breaking changes. the current mode of pinning the major version achieves the goal of not introducing breaking changes while also having a higher chance of bugs being handled automatically as their are found in the action code since we dont need to update for each minor and patch version bump.

PR #540, opened 2026-04-06
POC of a dynamic loading system for encoders to save on flash size

GL-PROXY:
# What's new

- Added lazy-loading support for SubGHz encoders
- Encoders can now be loaded from the SD card instead of being compiled directly into the firmware to save on flash size
- Currently only Security+ 2.0 has been finished and tested

# Overview

The goal is to help with the limited flash space by moving most of the encoder parts to the SD card and loading them dynamically when needed instead of keeping everything in flash. Currently this is more of a POC then a finished improvement there are many parts i could have done better and cut down on to save even more flash but i don't really have the time so i am just making this pr as a example of what could be done to help.

-----
# For the reviewer

- [X] I've uploaded the firmware with this patch to a device and verified its functionality
- [X] I've confirmed the bug to be fixed / feature to be stable

### DIFF
--- a/applications/main/subghz/application.fam
+++ b/applications/main/subghz/application.fam
@@ -61,3 +61,13 @@ App(
     # sources=["subghz_extended_freq.c"],
     order=650,
 )
+
+App(
+    appid="subghz_encoder_secplus_v2",
+    name="Security+ 2.0 Encoder",
+    apptype=FlipperAppType.PLUGIN,
+    entry_point="secplus_v2_encoder_plugin_ep",
+    targets=["f7"],
+    requires=["subghz"],
+    sources=["plugins/encoders/secplus_v2_encoder.c"],
+)
--- a/applications/main/subghz/helpers/subghz_txrx.c
+++ b/applications/main/subghz/helpers/subghz_txrx.c
@@ -1,5 +1,6 @@
 #include "subghz_txrx_i.h" // IWYU pragma: keep
 
+#include <lib/subghz/protocols/subghz_encoder_plugin_manager.h>
 #include <lib/subghz/protocols/protocol_items.h>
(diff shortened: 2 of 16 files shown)


WillyJL:
interesting idea, thanks for the pr. would be nice to have some numbers here, i have a feeling that most encoders rely on the same helpers functions as the decoders and if thats the case moving the encoders only wouldnt remove the bulk of the code from flash. i dont have numbers to back that up but thats something to test before moving forward with this

GL-PROXY:
> interesting idea, thanks for the pr. would be nice to have some numbers here, i have a feeling that most encoders rely on the same helpers functions as the decoders and if thats the case moving the encoders only wouldnt remove the bulk of the code from flash. i dont have numbers to back that up but thats something to test before moving forward with this

Hi WillyJL,

I’ll try to gather some numbers for you, though I might need to do it later since I’m currently busy. I’ll also look into which protocols are feasible, as you mentioned, it might not be possible for all of them.

GL-PROXY:
Hi WillyJL, here are the numbers you asked for.

Baseline (no patch):
.text         698128 (681.77 K)
.rodata       187892 (183.49 K)
.data            876 (  0.86 K)
.bss            4996 (  4.88 K)
.free_flash   161344 (157.56 K)

With lazy-loading patch (ONLY SECPLUSV2):
.text         697148 (680.81 K)
.rodata       188100 (183.69 K)
.data            876 (  0.86 K)
.bss            4996 (  4.88 K)
.free_flash   162112 (158.31 K)

WillyJL:
thanks for the numbers. not bad, but definitely diminishing for the amount of work to convert the protocols. ill keep this open as a draft and possibly come back to it later. thank you

PR #542, opened 2026-04-13
fix(APP): loading perf changes

kalicyh:
## Summary

This ports the `momentum_app` loading optimization without bringing in any localization changes.

## Changes

- lazy-load `mainmenu_app_labels` and `mainmenu_app_exes` only when entering the related scene
- lazy-load `asset_pack_names` only when entering the asset pack scene
- release both cached lists when no longer needed
- avoid building large VariableItemList/Submenu content until the corresponding subpage is actually opened
- keep UI strings unchanged from the target branch

## Why

Opening `momentum_app` was doing too much work up front:
- scanning asset packs during app alloc
- building menu app label/exe lists eagerly
- keeping large temporary lists alive longer than necessary

This reduces startup cost and lowers peak memory pressure when opening the app.

## Notes

This PR intentionally excludes CN/localization changes and only keeps the loading/perf behavior changes.

### DIFF
--- a/applications/main/momentum_app/momentum_app.c
+++ b/applications/main/momentum_app/momentum_app.c
@@ -210,7 +210,61 @@ void momentum_app_push_mainmenu_app(MomentumApp* app, FuriString* exe) {
     furi_string_free(label);
 }
 
+void momentum_app_load_asset_pack_names(MomentumApp* app) {
+    if(app->asset_pack_names_loaded) {
+        return;
+    }
+
+    app->asset_pack_index = 0;
+    File* folder = storage_file_alloc(app->storage);
+    FileInfo info;
+    char* name = malloc(ASSET_PACKS_NAME_LEN);
+    if(storage_dir_open(folder, ASSET_PACKS_PATH)) {
+        while(storage_dir_read(folder, &info, name, ASSET_PACKS_NAME_LEN)) {
+            if(info.flags & FSF_DIRECTORY && name[0] != '.') {
+                char* copy = strdup(name);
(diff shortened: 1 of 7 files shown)


kalicyh:
I've implemented the ability to freely switch between Chinese fonts in my Chinese firmware, and the firmware size looks decent at 880,473 bytes, leaving 167 bytes of free space compared to the previous 880,640 bytes.

This submission addresses a problem I encountered and resolved while developing the Chinese firmware. They were experiencing issues with the mntm application taking too long to open and even freezing, so I made the above changes.

If you also think it's good, you can adopt it.

WillyJL:
> Why?

you explained the effect it has, but not why it was needed. no one has ever complained of opening the momentum app being slow or having other issues to my knowledge

also, we explicitly do not want AI contributions

WillyJL:
> They were experiencing issues with the mntm application taking too long to open and even freezing, so I made the above changes.

this to me rather sounds like the AI that hallucinated some goofy way of doing this:

> I've implemented the ability to freely switch between Chinese fonts in my Chinese firmware

made many mistakes and created an issue that didnt exist before, and this was a workaround

kalicyh:
Can you give me some suggestions on how to optimize it?
CleanShot 2026-04-14 at 08 20 54@2x
CleanShot 2026-04-14 at 08 21 27@2x

https://github.com/kalicyh/Momentum-Firmware-CN/commit/5d1f0ad25596112db711faf8cc7da03f70524353

The modifications may not have been perfect, but they were effective.

PR #549, opened 2026-05-21
Clarify that making low-level things with AI is a bad idea.

GameLord2011:
# What's new

- I made sure that the AI is notified that this is a low-level project and that it is a bad idea to vibecode those

## Possible TODOs

- Notify the user that messing with low-level stuff and having no idea what you're doing at all is a bad idea? ~(Someone else should probably add this so that I'm not being hypocritical.)~
- Uhh should I have modified that rights-waivering notice?

-----
# For the reviewer

- [ ] I've uploaded the firmware with this patch to a device and verified its functionality [UNNEEDED AS THIS FALLS UNDER DOCUMENTATION-RELATED]
- [ ] I've confirmed the bug to be fixed / feature to be stable [DITTO]

### DIFF
--- a/.cursor/rules/momentum.mdc
+++ b/.cursor/rules/momentum.mdc
@@ -29,6 +29,7 @@ rules:
       - Preservation of software development as a human craft
       - Environmental responsibility regarding AI compute resources
       - Respect for maintainer time and expertise
+      - The fact that it is, in general, a bad idea to vibecode any type of device firmware
       
       Human contributors must:
       - Disable all AI coding assistants (GitHub Copilot, Cursor, etc.)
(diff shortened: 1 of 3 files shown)


Anonymous3-a:
s/furthur/further/

Also, didn't the agents file already make the AI refuse to work?

GameLord2011:
Yea, but it's probably a good idea to notify less-cooprative models, as they may take it as a suggestion.

Anonymous3-a:
Have you tested that hypothesis? I say that unless that is true, the agents file should be kept as-is, lest changes disturb the balance of weights

GameLord2011:
I'll convert ts to a draft until I can test it.

GameLord2011:
Actually nvm ts dumb, sry for wasting time.

PR #554, opened 2026-06-08
NFC: Add Ventra plugin to Metroflip imports, mark legacy parser deprecated

hazardousvoltage:
# What's new

- Mark existing legacy Ventra-UL parser deprecated -- ported it to Metroflip
- Add ventra plugin to Metroflip imports

-----
# For the reviewer


hazardousvoltage:
Submitting to upstream Metroflip instead as suggested

